PR #1, opened 2022-01-12
Feature/add e2e tests

rejoshed:
*Issue #, if available:*

*Description of changes:*

create/copy cloud-config
```
[Global]
api-key    = xxx
secret-key = xxx
api-url    = http://172.16.0.1:8080/client/api
```
make manifests
make kind-cluster
IMG=localhost:5000/cluster-api-provider-cloudstack:latest make docker-build
IMG=localhost:5000/cluster-api-provider-cloudstack:latest make docker-push
CLOUDSTACK_B64ENCODED_SECRET=$(base64 -i cloud-config) GINKGO_FOCUS=PR-Blocking make run-e2e
kind delete clusters capi-test


maxdrib (commented):
A few additional comments:
1. Do we need any e2e tests for affinity groups? Account/domain?
2. Can we also add a test for upgrade?
3. On a higher level, how do we maintain these tests? What's the policy for adding a new e2e test?

Makefile
```diff
@@ -43,10 +43,6 @@ all: build
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-MANIFEST_GEN_INPUTS=$(shell find ./api ./controllers -type f -name "*test*" -prune -o -name "*zz_generated*" -prune -o -print)
```
rejoshed, line 46: Where'd this go? We still need the inputs.

test/e2e/.gitignore
```diff
@@ -0,0 +1,2 @@
+kind-install-for-capd.sh
```
rejoshed, line 1: Why is this needed?

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+## --------------------------------------
+## Binaries
+## --------------------------------------
+
+BIN_DIR := $(REPO_ROOT)/bin
+GINKGO := $(BIN_DIR)/ginkgo
+KUSTOMIZE := $(BIN_DIR)/kustomize
+CONTROLLER_GEN := $(BIN_DIR)/controller-gen
+
+$(GINKGO): # Build ginkgo from bin folder.
```
rejoshed, line 25: Can we consolidate the binary building by just having the main makefile build these?

The main makefile can actually call this one, and in that target, we can have binaries as a prereq for e2e tests.

It would be nice too as we wouldn't need to specify the directory when calling make, so `make <target>` instead of `make --directory=./test/e2e <target>`

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+$(KUSTOMIZE): # Build kustomize from bin folder.
+	@mkdir -p $(BIN_DIR)
+	cd $(BIN_DIR) && curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
+
+$(CONTROLLER_GEN): ## Install controller-gen to bin.
+	@mkdir -p $(BIN_DIR)
+	GOBIN=$(BIN_DIR) go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.9
+
+## --------------------------------------
+## Kind Cluster
```
rejoshed, line 38: Same here. The main makefile already has a target for kind cluster.

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+	chmod +x kind-install-for-capd.sh
+	-./kind-install-for-capd.sh
+
+## --------------------------------------
+## Manifests
+## --------------------------------------
+
+.PHONY: manifests
+manifests: 
+	d=$$(date +%s) \
```
rejoshed, line 54: Same here... Main Makefile already has this target.

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+	$(KUSTOMIZE) build --load-restrictor LoadRestrictionsNone $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-invalid-worker-offering > $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-invalid-worker-offering.yaml
+	$(KUSTOMIZE) build --load-restrictor LoadRestrictionsNone $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-insufficient-compute-resources > $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-insufficient-compute-resources.yaml
+	$(KUSTOMIZE) build --load-restrictor LoadRestrictionsNone $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-node-drain > $(CLOUDSTACK_TEMPLATES)/v1beta1/cluster-template-node-drain.yaml
+
+## --------------------------------------
+## Docker build and push to local registry
+## --------------------------------------
+
+.PHONY: docker-build-and-push
+docker-build-and-push:
```
rejoshed, line 86: This is useful enough it would be nice to place it in the main makefile and call it something like `local-build-and-push`.

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+## End-to-End Tests
+## --------------------------------------
+
+.PHONY: e2e-essentials
+e2e-essentials: $(GINKGO) kind-cluster manifests cluster-templates docker-build-and-push
+
+.PHONY: run-e2e
+run-e2e: e2e-essentials ## Run End-to-End Tests
+	-d=$$(date +%s) \
+	&& PATH=$(BIN_DIR):${PATH} $(GINKGO) -v -trace -tags=e2e -skip=Conformance -nodes=1 --noColor=false ./... -- \
```
rejoshed, line 103: PATH can be set via export, and we actually do that from the main Makefile.

So, if you use the main Makefile to call this one, you shouldn't even need to set this path var.

`$(GINKGO)` in particular should just be `gingko` since it's now in `$PATH`

test/e2e/Makefile
```diff
@@ -0,0 +1,138 @@
...
+	    -e2e.artifacts-folder=${REPO_ROOT}/_artifacts \
+	    -e2e.config=${REPO_ROOT}/test/e2e/config/cloudstack.yaml \
+	    -e2e.skip-resource-cleanup=false -e2e.use-existing-cluster=true \
+	&& echo "This step took $$(($$(date +%s)-d)) seconds"
+	kind delete clusters capi-test
+
+.PHONY: run-e2e-pr-blocking
```
rejoshed, line 110: As far as I can tell, the only thing different between all these targets is the tag passed to ginkgo?

If so, can we just make the tag an arg the same way we run local jobs in the Prow repo?

[run job](https://github.com/aws/cluster-api-provider-cloudstack-prow-jobs/blob/cbe81fadfb11bc4431022df342e1fc11e792100a/Makefile#L47)

test/e2e/README.md
```diff
@@ -0,0 +1,64 @@
+# Testing
```
rejoshed, line 1: Do we need license headers here?

test/e2e/README.md
```diff
@@ -0,0 +1,64 @@
...
+|  `CLUSTER_ENDPOINT_PORT`                     | The cluster endpoint port                                                        | `6443`                      |
+|  `CLOUDSTACK_CONTROL_PLANE_MACHINE_OFFERING` | The machine offering for the control plane VM instances                          | `Large Instance`            |
+|  `CLOUDSTACK_WORKER_MACHINE_OFFERING`        | The machine offering for the worker node VM instances                            | `Medium Instance`           |
+|  `CLOUDSTACK_TEMPLATE_NAME`                  | The machine template for both control plane and worke node VM instances          | `kube-v1.20.10/ubuntu-2004` |
+|  `CLOUDSTACK_SSH_KEY_NAME`                   | The name of SSH key added to the VM instances                                    | `CAPCKeyPair6`              |
+
+You also have to export `CLOUDSTACK_B64ENCODED_SECRET` environment variable using this command `export CLOUDSTACK_B64ENCODED_SECRET=$(base64 -i cloud-config)` after creating `cloud-config` file with the following format.
```
rejoshed, line 36: Do we though? Seems we do that via the Makefile.

test/e2e/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The api-key and secret-key can be found or generated at Home > Accounts > admin > Users > admin of the ACS management UI. 
+
+### Running the e2e tests
+
+Run the following command to execute the CAPC e2e tests:
+
+```shell
+make run-e2e
```
rejoshed, line 52: Let's update this if we do change to using Makefile args.

test/e2e/data/fixture/expected-webpage.html
```diff
@@ -0,0 +1,14 @@
...
...oncealed in a most cunning disguisement, a desperate burglar slid into his happy home, and robbed them all of everything. And darker yet to tell, the blacksmith himself did ignorantly conduct this burglar into his family's heart. It was the Bottle Conjuror! Upon the opening of that fatal cork, forth flew the fiend, and shrivelled up his home. Now, for prudent, most wise, and economic reasons, the blacksmith's shop was in the basement of his dwelling, but with a separate entrance to it; so that always had the young and loving healthy wife listened with no unhappy nervousness, but with vigorous pleasure, to the stout ringing of her young-armed old husband's hammer; whose reverberations, muffled by passing through the floors and walls, came up to her, not unsweetly, in her nursery; and so, to stout Labor's iron lullaby, the blacksmith's infants were rocked to slumber. Oh, woe on woe! Oh, Death, why canst thou not sometimes be timely? Hadst thou taken this old blacksmith to thyself ere his full ruin came upon him, then had the young widow had a delicious grief, and her orphans a truly venerable, legendary sire to dream of in their after years; and all of them a care-killing competency.
```
rejoshed, line 10: > the young widow had a delicious grief, and her orphans a truly venerable, legendary sire to dream of in their after years; and all of them a care-killing competency.

Lulwhat? Haha

Makefile
wongni: Oops. I will put it back in

test/e2e/.gitignore
wongni: The file is downloaded from cluster-api github repo so I don't want to include it in our repo.

test/e2e/Makefile
wongni: I tried but I got this error when running e2e test from the main Makefile. 
```
Failed to compile e2e:

main module (github.com/aws/cluster-api-provider-cloudstack) does not contain package github.com/aws/cluster-api-provider-cloudstack/test/e2e
```
I think that Makefiles should be separated as well because test/e2e is a separate module.

test/e2e/Makefile
wongni: The reason why I didn't use kind-cluster target in the main Makefile is that it clones the entire cluster-api repo. In e2e test, it only needs to download kind-install-for-capd.sh file from the repo.

test/e2e/Makefile
rejoshed: This error is really just that pwd needs to be the e2e directory.

This can either be done via `cd && <test commands>`, or the main Makefile can call the e2e makefile.

Personally I prefer the one Makefile since I now understand it enough for it to know it can be a single small target.

test/e2e/Makefile
rejoshed: I don't think we care about that though do we? The time it takes to pull down the cluster-api repo is seconds at most, and not that much data.

To me fewer code paths is better than saving the few megabytes of data transfer here.

test/e2e/Makefile
rejoshed: > I think that Makefiles should be separated as well because test/e2e is a separate module.

By this reasoning, we should push this to be a separate repo even.

The one Makefile already manages several modules in this repo.

rejoshed: Also, in retrospect, the cd shouldn't be needed. Gingko will run in a specified directory. It just needs to be passed the directory.

Makefile
```diff
@@ -177,16 +186,36 @@ pkg/mocks/mock%.go: $(shell find ./pkg/cloud -type f -name "*test*" -prune -o -p
 ##@ Tilt
 
 .PHONY: tilt-up 
-tilt-up: cluster-api kind-cluster cluster-api/tilt-settings.json manifests cloud-config # Setup and run tilt for development.
+tilt-up: cluster-api kind-cluster cluster-api/tilt-settings.json manifests cloud-config ## Setup and run tilt for development.
```
rejoshed, line 189: Good catch!

Makefile
```diff
@@ -177,16 +186,36 @@ pkg/mocks/mock%.go: $(shell find ./pkg/cloud -type f -name "*test*" -prune -o -p
...
+##@ End-to-End Testing
+
+CLOUDSTACK_TEMPLATES := $(PROJECT_DIR)/test/e2e/data/infrastructure-cloudstack/v1beta1/
+CLUSTER_TEMPLATES_INPUTS=$(shell find $(CLOUDSTACK_TEMPLATES) -type d -name "cluster*")
+e2e-cluster-templates: bin/kustomize $(CLOUDSTACK_TEMPLATES)/.templateflag.mk ## Generate cluster template files for e2e testing.
+$(CLOUDSTACK_TEMPLATES)/.templateflag.mk: $(CLUSTER_TEMPLATES_INPUTS)
+	$(shell find $(CLOUDSTACK_TEMPLATES) -type d -name "cluster*" -exec bash -c "bin/kustomize build --load-restrictor LoadRestrictionsNone {} > {}.yaml" \;)
+	@touch $(CLOUDSTACK_TEMPLATES)/.templateflag.mk
+
+e2e-essentials: bin/ginkgo e2e-cluster-templates kind-cluster ## Fulfill essential tasks for e2e testing.
```
rejoshed, line 211: Can we just move these to dependencies of the e2e tests themselves? No need to have an extra target.

Makefile
```diff
@@ -80,6 +82,13 @@ bin/manager: $(MANAGER_BIN_INPUTS)
 	go vet ./...
 	go build -o bin/manager main.go
 
+.PHONY: build-for-docker
+build-for-docker: bin/manager-linux-amd64 ## Build manager binary for docker image building.
+bin/manager-linux-amd64: $(MANAGER_BIN_INPUTS)
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
```
jweite-amazon, line 88: I'd like to have more confidence that cross-platform building locally, ostensibly on darwin, doesn't introduce differences that actually building in linux (debian in this case) env doesn't.  I don't see other CAPI providers doing this.  Do you have any research that supports the safety/reliability of this technique?

rejoshed: I'm fine either way @wongni  if you do move back to the old way, we can just put this Dockerfile in hack and use it only for e2e.

wongni: I didn't do any research about how reliable golang's cross-compilation is, but when I tested a CAPC image built in this way on my Macbook in a Linux VM it worked well.

test/e2e/data/fixture/expected-webpage.html
wongni: Replacing it to `Lorem Ipsum`

PR #3, opened 2022-01-12
Feature/domain setting

rejoshed:
Issue # -- CTECH-94

**This PR was rebased upon the Tilt fixup branch. Once that is merged, it will not show those changes.** 

Description of changes:
If specified, resources will be created under account and domain ids.

Here a VM can be seen under my TestAccountName account which resides in my TestDomain.
image


api/v1alpha3/cloudstackcluster_webhook.go
```diff
@@ -54,19 +55,20 @@ var _ webhook.Validator = &CloudStackCluster{}
...
-	)
+	var errorList field.ErrorList
 
 	// IdentityRefs must be Secrets.
-	if spec.IdentityRef != nil && spec.IdentityRef.Kind != defaultIdentityRefKind {
+	if r.Spec.IdentityRef != nil && r.Spec.IdentityRef.Kind != defaultIdentityRefKind {
 		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
+	if (r.Spec.Account == "") != (r.Spec.Domain == "") {
```
rejoshed, line 65: Require both account and domain if set.

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/onsi/ginkgo v1.16.4
-	github.com/onsi/gomega v1.16.0
+	github.com/onsi/gomega v1.17.0
```
rejoshed, line 12: Bumped for better gomega matchers.

pkg/cloud/helpers_test.go
```diff
@@ -46,3 +49,27 @@ func getConfigPath(filename string) string {
 	dir, _ := os.Getwd()
 	return path.Join(dir, FixturePath, filename)
 }
+
+// This matcher is used to make gomega matching compatible with gomock parameter matching.
+// It's pretty awesome!
```
rejoshed, line 54: Like seriously!!!

pkg/cloud/instance_test.go
```diff
@@ -181,6 +182,49 @@ var _ = Describe("Instance", func() {
 			Ω(client.GetOrCreateVMInstance(csMachine, machine, csCluster, "")).Should(MatchError("unknown err"))
 		})
 
+		describeDomainAccountTest := func(desc string) func(string, string) string {
+			return func(account string, domainID string) string {
+				return fmt.Sprintf("%s and %s to beet to, %s ", account, domainID, desc)
```
jweite-amazon, line 187: beet?

api/v1alpha3/cloudstackcluster_webhook.go
maxdrib: isn't domain itself sufficient for performing all the necessary operations? Why do we also need account? Also is this an account ID? Or the account name?

rejoshed: Sorry, I worded it wrong.

If account is set, domain must be set too.

https://cloudstack.apache.org/api/apidocs-4.16/apis/deployVirtualMachine.html

It's a restriction imposed by the cloudstack api itself.

> 

account | an optional account for the virtual machine. Must be used with domainId. | false
-- | -- | --

maxdrib: right so wouldn't this check fail if only domain is provided, even though that's an acceptable situation?

PR #8, opened 2022-01-19
Feature/affinity management

rejoshed:
*Issue #: CTECH-71*

**Please consider my comments! There is a lot of detail in them.**

*Description of changes:*
Primarily adds affinity group management, but also fixes two issues with public endpoint setting. 1). We couldn't set the Endpoint from the controller because of webhooks. 2). The IP fetching was just grabbing the first IP in a list without properly checking it.

*Testing performed:*
Created a cluster 1:1 with 2 affinity groups.
Deleted said cluster.
Created a cluster 1:1 with 1 affinity group on the control plane.
Deleted said cluster.
Created a cluster 3:2 with 2 affinity groups.
Scaled down to 3:1
Scaled down to 1:1
Affinity groups still present.
Deleted the cluster.

Updated naming scheme.
image

image
image
image

image
image
image


Makefile
```diff
@@ -163,7 +163,7 @@ clean: ## Clean.
 export KUBEBUILDER_ASSETS=$(PROJECT_DIR)/bin
 
 .PHONY: test
```
rejoshed, line 165: If the mocks need updating, they need to be updated prior to linting and other commands.

Makefile
```diff
@@ -163,7 +163,7 @@ clean: ## Clean.
...
-test: lint generate-deepcopy generate-mocks bin/ginkgo bin/kubectl bin/kube-apiserver bin/etcd ## Run tests. At the moment this is only unit tests.
+test: generate-mocks lint generate-deepcopy bin/ginkgo bin/kubectl bin/kube-apiserver bin/etcd ## Run tests. At the moment this is only unit tests.
```
rejoshed, line 166: If the mocks need updating, they need to be updated prior to linting and other commands.

PROJECT
```diff
@@ -11,8 +11,8 @@ resources:
   domain: cluster.x-k8s.io
   group: infrastructure
   kind: CloudStackCluster
-  path: github.com/aws/cluster-api-provider-cloudstack/api/v1alpha3
-  version: v1alpha3
+  path: github.com/aws/cluster-api-provider-cloudstack/api/v1beta1
+  version: v1beta1
```
rejoshed, line 15: Some remaining v1beta1 bump outliers.

README.md
```diff
@@ -38,7 +38,9 @@ Install [tilt prerequisites](https://cluster-api.sigs.k8s.io/developer/tilt.html
 
 ## Running CAPC without Tilt - detailed instructions
 
-Generally speaking, this cloudstack infrastructure provider will generate a docker image and 3 yaml files. `clusterctl` (a binary tool) will use the above docker image and 3 yaml files to provision a cluster from your local machine using cloudstack as a provider.
+Generally speaking, this cloudstack infrastructure provider will generate a docker image and 3 yaml files. 
+`clusterctl` (a binary tool) will use the above docker image and 3 yaml files to provision a cluster from your local 
+machine using cloudstack as a provider.
```
rejoshed, line 43: Largely cleaned description to reflect v1beta1 and wrapped some very long lines.

api/v1beta1/cloudstackcluster_webhook.go
```diff
@@ -96,8 +96,12 @@ func (r *CloudStackCluster) ValidateUpdate(old runtime.Object) error {
...
 	errorList = webhook_utilities.EnsureStringFieldsAreEqual(spec.Zone, oldSpec.Zone, "zone", errorList)
 	errorList = webhook_utilities.EnsureStringFieldsAreEqual(spec.Network, oldSpec.Network, "network", errorList)
-	errorList = webhook_utilities.EnsureStringFieldsAreEqual(spec.ControlPlaneEndpoint.Host, oldSpec.ControlPlaneEndpoint.Host, "controlplaneendpointhost", errorList)
-	errorList = webhook_utilities.EnsureStringFieldsAreEqual(string(spec.ControlPlaneEndpoint.Port), string(oldSpec.ControlPlaneEndpoint.Port), "controlplaneendpointport", errorList)
+	if oldSpec.ControlPlaneEndpoint.Host != "" { // Need to allow one time endpoint setting via CAPC cluster controller.
+		errorList = webhook_utilities.EnsureStringFieldsAreEqual(
+			spec.ControlPlaneEndpoint.Host, oldSpec.ControlPlaneEndpoint.Host, "controlplaneendpointhost", errorList)
+		errorList = webhook_utilities.EnsureStringFieldsAreEqual(
+			string(spec.ControlPlaneEndpoint.Port), string(oldSpec.ControlPlaneEndpoint.Port), "controlplaneendpointport", errorList)
+	}
```
rejoshed, line 104: This is a fixup for the scenario in which the host or port is not specified at cluster creation time.

The cluster controller needs to be able to set this.

Makefile
```diff
@@ -56,7 +56,7 @@ config/.flag.mk: bin/controller-gen $(MANIFEST_GEN_INPUTS)
 
 .PHONY: release-manifests
 RELEASE_MANIFEST_TARGETS=$(RELEASE_DIR)/infrastructure-components.yaml $(RELEASE_DIR)/metadata.yaml
-RELEASE_MANIFEST_INPUTS=config/.flag.mk $(shell find config)
+RELEASE_MANIFEST_INPUTS=bin/kustomize config/.flag.mk $(shell find config)
```
rejoshed, line 59: Was missing bin/kustomize as a pre-req here. Thanks @wongni for finding this!

api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -55,18 +57,25 @@ var _ webhook.Validator = &CloudStackMachine{}
...
-	errorList = webhook_utilities.EnsureFieldExists(spec.Template, "Template", errorList)
+	affinity := strings.ToLower(r.Spec.Affinity)
+	if !(affinity == "" || affinity == "no" || affinity == "pro" || affinity == "anti") {
+		errorList = append(errorList, field.Invalid(field.NewPath("spec", "Affinity"), r.Spec.Affinity,
+			`Affinity must be "no", "pro", "anti", or unspecified.`))
+	}
+	if (affinity != "no") && affinity != "" && (len(r.Spec.AffinityGroupIds) > 0) {
+		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "AffinityGroupIds"),
+			"AffinityGroupIds cannot be specified when Affinity is specified as anything but `no`"))
+	}
```
rejoshed, line 75: Input argument validation.

config/rbac/role.yaml
```diff
@@ -39,6 +38,7 @@ rules:
   resources:
   - machines
   - machines/status
+  - machinesets
```
rejoshed, line 41: In order to know when an affinity group needs deletion, the machine controller needs to watch for the deletion of its management owner.

In the case of workload nodes, that's the machineset pointed to by the CAPI machine that owns the CAPC machine.

Adding this permission allows the machine controller to list machinesets to check for deletion thereof.

controllers/cloudstackcluster_controller.go
```diff
@@ -21,6 +21,8 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	"github.com/hashicorp/go-multierror"
```
rejoshed, line 24: We had a few issues in error handling patches that made troubleshooting the endpoint/webhook issue difficult.

Instead of getting tricky about returning errors, I instead use multierror and return both errors.

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 	// Get CloudStack cluster.
 	csCluster := &infrav1.CloudStackCluster{}
-	if retErr = r.Client.Get(ctx, req.NamespacedName, csCluster); retErr != nil {
-		if client.IgnoreNotFound(retErr) == nil {
+	if err := r.Client.Get(ctx, req.NamespacedName, csCluster); err != nil {
```
rejoshed, line 74: Named vars can be tricky.

In our case, retErr is actually modified by the patch defer func if an error occurs during patching.

As such, we need to be careful not to re-declare retErr or the named return var functionality is lost.

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		if client.IgnoreNotFound(err) == nil {
 			log.Info("Cluster not found.")
 		}
-		return ctrl.Result{}, client.IgnoreNotFound(retErr)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	// Get CAPI cluster.
-	cluster, retErr := util.GetOwnerCluster(ctx, r.Client, csCluster.ObjectMeta)
```
rejoshed, line 80: This is a re-declaration, and the error handling in the defer is thus lost.

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	if retErr != nil {
-		return reconcile.Result{}, retErr
-	} else if cluster == nil {
+	capiCluster, err := util.GetOwnerCluster(ctx, r.Client, csCluster.ObjectMeta)
```
rejoshed, line 82: I've decided I like the naming scheme of cs for cloudstack objects, and capi for capi objects.

In this case it clarifies quite a bit. Instead of `csCluster` & `cluster` we have `csCluster` & capiCluster`.

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	if patchHelper, err := patch.NewHelper(csCluster, r.Client); err != nil {
+		return ctrl.Result{}, err
 	} else {
 		defer func() {
-			if err := patchHelper.Patch(ctx, csCluster); retErr == nil && err != nil {
-				retErr = err
+			if err = patchHelper.Patch(ctx, csCluster); err != nil {
+				msg := "error patching CloudStackCluster %s/%s"
+				err = errors.Wrapf(err, msg, csCluster.Namespace, csCluster.Name)
+				retErr = multierror.Append(retErr, err)
```
rejoshed, line 104: Yay! Sane error handling!

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 			}
 		}()
 	}
 
 	// Delete Cluster Resources if deletion timestamp present.
 	if !csCluster.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(log, csCluster)
+		return r.reconcileDelete(ctx, log, capiCluster, csCluster)
```
rejoshed, line 111: I did some reading, and apparently context objects should always be the first argument in a function/method, and they should not be jammed into structs.

So, it's the first arg here.

Still, I'd like us to transition from passing so many args (not that bad in the cluster case, but the machine can get unwieldy)  to passing structs.

In this case it'd be something like `r.reconcileDelete(ctx, clstrCtrlrScope)`, but this conversion will be a bit of a chore to complete.

controllers/cloudstackcluster_controller_test.go
```diff
@@ -77,6 +77,8 @@ var _ = Describe("CloudStackClusterReconciler", func() {
 
 	It("Should create a cluster", func() {
 		By("Fetching a CS Cluster Object")
+		CS.EXPECT().GetOrCreateCluster(gomock.Any()).MinTimes(1)
```
rejoshed, line 80: This may actually have been our random test failure issue.

I think I was just wrong...

I ran the test locally and saw a failure after only ~15 seconds. So, the failure wasn't timeout based.

I then realize it was actually complaining about a method call that happened -- not a missing call.

Turns out the call can sometimes happen early.

controllers/cloudstackmachine_controller.go
```diff
@@ -203,8 +207,18 @@ func (r *CloudStackMachineReconciler) reconcile(
...
+	csMachine *infrav1.CloudStackMachine,
+	capiMachine *capiv1.Machine,
+	csCluster *infrav1.CloudStackCluster) (ctrl.Result, error) {
+
+	// Remove any CAPC managed Affinity groups if owner references a deleted object.
+	if csCtrlrUtils.IsOwnerDeleted(ctx, r.Client, capiMachine) {
+		if err := r.RemoveManagedAffinity(log, capiMachine, csMachine, csCluster); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
rejoshed, line 221: This is the crux of the managed affinity deletion work.

controllers/cloudstackmachine_controller_utils.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright 2022.
```
rejoshed, line 2: The machine controller file was getting bloated.

Broke some of the methods out here.

controllers/utils/utils.go
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2022.
```
rejoshed, line 2: Both the cloud package and the controllers package need these utils, so I put them in their own package to avoid circular imports.

Now that I understand all this stuff better, I no longer believe the pkg/cloud separation is the best pattern to organize by, but here we are. I doubt I'll get to it, it's definitely not a priority, but it would be nice to rethink and reorg.

pkg/cloud/affinity_groups_test.go
```diff
@@ -0,0 +1,119 @@
...
+		ags.EXPECT().GetAffinityGroupByID(fakeAG.Id).Return(nil, -1, errors.New("FakeError"))
+		ags.EXPECT().NewCreateAffinityGroupParams(fakeAG.Name, fakeAG.Type).
+			Return(&cloudstack.CreateAffinityGroupParams{})
+		ags.EXPECT().CreateAffinityGroup(ParamMatch(And(AccountEquals("FakeAccount"), DomainIdEquals("FakeDomainId")))).
+			Return(&cloudstack.CreateAffinityGroupResponse{}, nil)
+
+		Ω(client.GetOrCreateAffinityGroup(cluster, fakeAG)).Should(Succeed())
+	})
+
+	Context("AffinityGroup Integ Tests", func() {
```
rejoshed, line 82: I feel pretty cool about this test setup.

Essentially, the integ suite is just skipped if there is no cloud config to work with. This way I can run integ tests locally, and eventually against our ACS setup too, without breaking things in the meantime.

pkg/cloud/client.go
```diff
@@ -23,27 +23,26 @@ import (
...
 //go:generate mockgen -destination=../mocks/mock_client.go -package=mocks github.com/aws/cluster-api-provider-cloudstack/pkg/cloud Client
 
 type Client interface {
-	GetOrCreateCluster(*infrav1.CloudStackCluster) error
-	GetOrCreateVMInstance(*infrav1.CloudStackMachine, *capiv1.Machine, *infrav1.CloudStackCluster, string) error
-	ResolveVMInstanceDetails(*infrav1.CloudStackMachine) error
-	DestroyVMInstance(*infrav1.CloudStackMachine) error
-	AssignVMToLoadBalancerRule(*infrav1.CloudStackCluster, string) error
+	ClusterIface
+	VMIface
```
rejoshed, line 32: Slowly moving to use individual interfaces.

pkg/cloud/client.go
```diff
@@ -23,27 +23,26 @@ import (
...
 	ResolvePublicIPDetails(*infrav1.CloudStackCluster) (*cloudstack.PublicIpAddress, error)
 	ResolveLoadBalancerRuleDetails(*infrav1.CloudStackCluster) error
 	GetOrCreateLoadBalancerRule(*infrav1.CloudStackCluster) error
+	AffinityGroupIFace
 }
 
 type client struct {
 	cs *cloudstack.CloudStackClient
+	// This is a placeholder for sending non-blocking requests.
+	// csA *cloudstack.CloudStackClient
```
rejoshed, line 45: I wanted to leave this in here. When we want to send a delete command without waiting, this is how we'll do it.

pkg/cloud/instance.go
```diff
@@ -159,6 +167,23 @@ func (c *client) GetOrCreateVMInstance(
 
 	if len(csMachine.Spec.AffinityGroupIds) > 0 {
 		p.SetAffinitygroupids(csMachine.Spec.AffinityGroupIds)
+	} else if !(strings.ToLower(csMachine.Spec.Affinity) == "no" || csMachine.Spec.Affinity == "") {
+		ownerRef := csCtrlrUtils.GetManagementOwnerRef(machine)
```
rejoshed, line 171: Managed affinity creation. : )

pkg/cloud/instance_test.go
```diff
@@ -191,34 +191,18 @@ var _ = Describe("Instance", func() {
...
-					vms.EXPECT().GetVirtualMachinesMetricByID(*csMachine.Spec.InstanceID).
-						Return(&cloudstack.VirtualMachinesMetric{}, 1, nil))
-
-				vms.EXPECT().NewDeployVirtualMachineParams(serviceOfferingID, templateID, csCluster.Status.ZoneID).
-					Return(&cloudstack.DeployVirtualMachineParams{})
-
 				csCluster.Spec.Account = account
 				csCluster.Status.DomainID = domainID
-				deploymentResp := &cloudstack.DeployVirtualMachineResponse{Id: *csMachine.Spec.InstanceID}
-				accountMatcher := WithTransform( // Call GetAccount on th DeployVM param passed.
```
rejoshed, line 209: These are now handled by the matcher I created above.

pkg/cloud/network.go
```diff
@@ -89,21 +89,33 @@ func (c *client) GetOrCreateNetwork(csCluster *infrav1.CloudStackCluster) (retEr
...
-	if ip := csCluster.Spec.ControlPlaneEndpoint.Host; ip != "" {
+	if ip != "" {
 		p.SetIpaddress(ip)
 	}
 	publicAddresses, err := c.cs.Address.ListPublicIpAddresses(p)
+
 	if err != nil {
 		return nil, err
-	} else if publicAddresses.Count > 0 {
+	} else if ip != "" && publicAddresses.Count == 1 { // Endpoint specified and IP found.
```
rejoshed, line 105: This is a fix that actually checks for public IP address availability.

Previously we would just the first IP address returned.

Makefile
jweite-amazon: Why is this ordering critical, particularly wrt lint, which I imagine to be read-only.

controllers/cloudstackcluster_controller.go
jweite-amazon: I'm kinda on the other side of that argument.  I find listing the params individually in the function signature lets me see at a glimpse what's coming in, promoting grok-ability.  Yes, it's a bit more work to code/maintain, but I find it to be worth it.  

So I'm not a fan of *artificially* gathering params into structs.  (Of course, structs that define the objects of the app should be passed in as structs...)

controllers/cloudstackcluster_controller.go
```diff
@@ -69,52 +71,58 @@ func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	return r.reconcile(log, csCluster)
+	return r.reconcile(ctx, log, capiCluster, csCluster)
 }
 
 // Actually reconcile cluster.
 func (r *CloudStackClusterReconciler) reconcile(
+	ctx context.Context,
 	log logr.Logger,
-	csCluster *infrav1.CloudStackCluster) (ctrl.Result, error) {
+	capiCluster *capiv1.Cluster,
```
jweite-amazon, line 122: Why adding this unused param?  (Also in reconcileDelete).

controllers/cloudstackmachine_controller.go
```diff
@@ -70,93 +74,95 @@ func (r *CloudStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 		} else {
-			return ctrl.Result{}, retErr
+			return ctrl.Result{}, err
 		}
 	} else if csCluster.Status.ZoneID == "" {
 		log.Info("CloudStackCluster ZoneId not initialized. Likely not ready.")
 		return ctrl.Result{RequeueAfter: RequeueTimeout}, nil
 	}
 
+	// Delete VM instance if deletion timestamp present.
```
jweite-amazon, line 138: This "delete VM" code block was moved above the "Fetch Cloudstack Cluster" block in https://github.com/aws/cluster-api-provider-cloudstack/commit/2794cc800aad7fa206b21444d6bd6802c8f7408d to address a bug where machine deletion fails if cluster has already been deleted.

templates/cluster-template-managed-ssh.yaml
```diff
@@ -29,7 +29,7 @@ spec:
   network: ${CLOUDSTACK_NETWORK_NAME}
   zone: ${CLOUDSTACK_ZONE_NAME}
 ---
-apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
 kind: KubeadmControlPlane
```
jweite-amazon, line 33: KubeAdmControlPlane changed slightly from v1alpha3 to v1alpha4.  See [Required cluster template changes](https://cluster-api.sigs.k8s.io/developer/providers/v0.3-to-v0.4.html#required-cluster-template-changes).  infrastructureTemplate becomes machineTemplate.infrastructureRef.

rejoshed:
Fix slashes. "Deleting affinity group \"Affinity-4112d34f-908c-4c59-848d-ebbed985d96c\"" "machi

Makefile
rejoshed: The mocks are used in the tests. If the mock clients don't match the interface anymore as the actual has changed then `go fmt` and `go vet` break as the test code just wouldn't even compile.

templates/cluster-template-managed-ssh.yaml
rejoshed: Good catch! I need to get to where I actually use the templates directly so I notice these issues. Right now I've been maintaining my own version outside the repo. Will update shortly.

api/v1beta1/cloudstackcluster_webhook.go
maxdrib: why wouldn't host/port be specified at cluster creation time?

api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -55,18 +57,25 @@ var _ webhook.Validator = &CloudStackMachine{}
...
 	}
 
-	errorList = webhook_utilities.EnsureFieldExists(spec.Offering, "Offering", errorList)
-	errorList = webhook_utilities.EnsureFieldExists(spec.Template, "Template", errorList)
+	affinity := strings.ToLower(r.Spec.Affinity)
+	if !(affinity == "" || affinity == "no" || affinity == "pro" || affinity == "anti") {
+		errorList = append(errorList, field.Invalid(field.NewPath("spec", "Affinity"), r.Spec.Affinity,
+			`Affinity must be "no", "pro", "anti", or unspecified.`))
+	}
+	if (affinity != "no") && affinity != "" && (len(r.Spec.AffinityGroupIds) > 0) {
```
maxdrib, line 72: some inconsistency with the parentheses here which would be good to clean up. Alternately, can we specify these attributes as mutually exclusive, as in https://github.com/go-playground/validator/issues/608#issuecomment-956253612?

api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -55,18 +57,25 @@ var _ webhook.Validator = &CloudStackMachine{}
...
 	// IdentityRefs must be Secrets.
-	if spec.IdentityRef != nil && spec.IdentityRef.Kind != defaultIdentityRefKind {
+	if r.Spec.IdentityRef != nil && r.Spec.IdentityRef.Kind != defaultIdentityRefKind {
 		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
-	errorList = webhook_utilities.EnsureFieldExists(spec.Offering, "Offering", errorList)
-	errorList = webhook_utilities.EnsureFieldExists(spec.Template, "Template", errorList)
+	affinity := strings.ToLower(r.Spec.Affinity)
+	if !(affinity == "" || affinity == "no" || affinity == "pro" || affinity == "anti") {
```
maxdrib, line 68: Can we just define enums for these instead of using strings? https://www.sohamkamani.com/golang/enums/#enum-types

api/v1beta1/cloudstackmachine_webhook_test.go
```diff
@@ -231,7 +231,8 @@ var _ = Describe("CloudStackMachine webhook", func() {
 
 			cloudStackMachine.DeepCopyInto(cloudStackMachineUpdate)
 			cloudStackMachineUpdate.Spec.AffinityGroupIds = []string{"28b907b8-75a7-4214-bd3d-6c61961fc2af"}
-			Expect(k8sClient.Update(ctx, cloudStackMachineUpdate)).Should(Succeed())
+			Expect(k8sClient.Update(ctx, cloudStackMachineUpdate).Error()).Should(MatchRegexp(forbiddenRegex, "AffinityGroupIds"))
+			// Expect(k8sClient.Update(ctx, cloudStackMachineUpdate)).Should(Succeed())
```
maxdrib, line 235: should this commented line be removed?

controllers/cloudstackmachine_controller.go
```diff
@@ -70,93 +74,95 @@ func (r *CloudStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-	return r.reconcile(log, ctx, csCluster, csMachine, machine)
+	return r.reconcile(ctx, log, csMachine, capiMachine, csCluster)
 }
 
 // Actually reconcile/Create a VM instance.
 func (r *CloudStackMachineReconciler) reconcile(
-	log logr.Logger,
 	ctx context.Context,
-	csCluster *infrav1.CloudStackCluster,
+	log logr.Logger,
```
maxdrib, line 150: why are we passing this around? Should it just be saved at the Reconciler level and referenced as r.log?

controllers/cloudstackmachine_controller.go
```diff
@@ -70,93 +74,95 @@ func (r *CloudStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	log.V(1).Info("reconcile CloudStackMachine")
 
 	// Make sure bootstrap data is available in CAPI machine.
-	if machine.Spec.Bootstrap.DataSecretName == nil {
+	if capiMachine.Spec.Bootstrap.DataSecretName == nil {
 		log.Info("Bootstrap DataSecretName not yet available.")
 		return ctrl.Result{}, nil
 	}
-	log.Info("Got Bootstrap DataSecretName: " + *machine.Spec.Bootstrap.DataSecretName)
+	log.Info("Got Bootstrap DataSecretName.")
```
maxdrib, line 162: why make this log less informative? Is it a security issue, or just spammy?

controllers/cloudstackmachine_controller.go
```diff
@@ -203,8 +209,20 @@ func (r *CloudStackMachineReconciler) reconcile(
...
+	csMachine *infrav1.CloudStackMachine,
+	capiMachine *capiv1.Machine,
+	csCluster *infrav1.CloudStackCluster) (ctrl.Result, error) {
+
+	// Remove any CAPC managed Affinity groups if owner references a deleted object.
+	if deleted, err := csCtrlrUtils.IsOwnerDeleted(ctx, r.Client, capiMachine); err != nil {
+		return ctrl.Result{}, err
+	} else if deleted {
+		if err := r.RemoveManagedAffinity(log, capiMachine, csMachine, csCluster); err != nil {
+			return ctrl.Result{}, err
```
maxdrib, line 223: do we want to add any additional explanation of where the error happened and what might've occurred, instead of just returning `err`?

controllers/cloudstackmachine_controller_utils.go
maxdrib: I'm not sure I agree with splitting out methods for the same object across multiple files. That'll make it more confusing to navigate and test, I feel, especially for a new maintainer. I'm inclined to keep all the relevant code in a single file instead of hiding some functions

controllers/utils/utils.go
```diff
@@ -0,0 +1,128 @@
...
+	if util.IsControlPlaneMachine(capiMachine) {
+		// The controlplane sticks around after deletion pending the deletion of its machiens.
+		// As such, need to check the deletion timestamp thereof.
+		if cp, err := GetKubeadmControlPlaneFromCAPIMachine(ctx, client, capiMachine); cp != nil && cp.DeletionTimestamp == nil {
+			return false, nil
+		} else if err != nil && !strings.Contains(err.Error(), "not found") {
+			return false, err
+		}
+	} else {
+		// The machienset is deleted immediately, regardless of machine ownership.
```
maxdrib, line 100: nit: typo in machineset*

controllers/utils/utils.go
```diff
@@ -0,0 +1,128 @@
...
+			return false, nil
+		} else if err != nil && !strings.Contains(err.Error(), "not found") {
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+// fetchRef simply searches a list of OwnerReference objects for a given kind.
+func fetchRef(reflist []meta.OwnerReference, kind string) *meta.OwnerReference {
```
maxdrib, line 112: nit:fetchOwnerRef*? And/or capitalize the L in refList for proper camel casing

controllers/utils/utils_test.go
```diff
@@ -0,0 +1,42 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	// . "github.com/onsi/ginkgo/extensions/table"
+	// . "github.com/onsi/gomega"
```
maxdrib, line 22: nit: remove unused imports. And what's the plan for this file? Will it be done in this PR? Or separately?

pkg/cloud/affinity_groups_test.go
maxdrib: How do we avoid unintentionally skipping these tests?

pkg/cloud/client.go
maxdrib: whoa, this looks interesting. Can you explain what wizardry is happening here?

controllers/cloudstackmachine_controller_utils.go
jweite-amazon: I personally don't consider module size (in lines of code) to be particularly sinful.  The current 317 lines plus the 77 from this module don't concern me.

The fact that RemoveManagedAffinity is only called from reconcileDelete() in the machine controller also dissuades me from breaking it out into a utility module.  (I'd usually put shared functions there).

I don't see any calls to ResolveManagedAffinity.

controllers/utils/utils_test.go
```diff
@@ -0,0 +1,42 @@
+/*
```
jweite-amazon, line 1: If this test module is present, I think it should implement tests.  Otherwise I'd leave it out.

pkg/cloud/client.go
maxdrib: I'm not a huge fan of leaving comments to help with future work, especially minor things like syntax. I'd be inclined to remove, especially since there's no timeline for this work actually being done

controllers/utils/utils_test.go
rejoshed: Whoops... Didn't mean to add that in just yet.

pkg/cloud/affinity_groups_test.go
```diff
@@ -0,0 +1,119 @@
...
+		machine.ObjectMeta.SetName("rejoshed-affinity-group-test-vm")
+		capiMachine = &capiv1.Machine{}
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
+	})
+
+	It("fetches an affinity group", func() {
+		ags.EXPECT().GetAffinityGroupByName(fakeAG.Name).Return(&cloudstack.AffinityGroup{}, 1, nil)
```
jweite-amazon, line 65: These cover happy day GetOrCreateAffinityGroup. What's the plan for the other functions in affinity_groups.go?

jweite-amazon: (Or do the integration tests below effectively cover the rest?)

pkg/cloud/instance.go
```diff
@@ -159,6 +167,22 @@ func (c *client) GetOrCreateVMInstance(
 
 	if len(csMachine.Spec.AffinityGroupIds) > 0 {
 		p.SetAffinitygroupids(csMachine.Spec.AffinityGroupIds)
+	} else if !(strings.ToLower(csMachine.Spec.Affinity) == "no" || csMachine.Spec.Affinity == "") {
```
jweite-amazon, line 170: While logically correct, I glossed-over the leading ! while reading, and then wondered why you were creating affinity groups when the param is "no" or blank.  Took me a head-scratch or two to find the !.  For readability I prefer:

if (strings.ToLower(csMachine.Spec.Affinity) != "no" && csMachine.Spec.Affinity != "")

It's reads closer to how I'd say it too: if it's not no and it's not blank.

pkg/cloud/instance.go
```diff
@@ -159,6 +167,22 @@ func (c *client) GetOrCreateVMInstance(
...
+		ownerRef := csCtrlrUtils.GetManagementOwnerRef(machine)
+		if ownerRef == nil {
+			return errors.Errorf("Could not find management owner reference for %s/%s",
+				csMachine.Namespace, csMachine.Name)
+		}
+		name := fmt.Sprintf("Affinity-%s", ownerRef.UID)
```
jweite-amazon, line 176: Following from today's demo, can prefix the name with something descriptive, to make the AG's role clearer for the admin or troubleshooter?

pkg/cloud/instance.go
```diff
@@ -159,6 +167,22 @@ func (c *client) GetOrCreateVMInstance(
...
+		affinityType := AffinityGroupType
+		if strings.ToLower(csMachine.Spec.Affinity) == "anti" {
```
jweite-amazon, line 178: Following from Rob's previous feedback, let's avoid literals in code.

rejoshed:
@maxdrib 

> whoa, this looks interesting. Can you explain what wizardry is happening here?

It's interface composition. Just breaks out the individual interfaces so they're easier to define and manage.

It's not too different from struct composition. The various interfaces are defined in their own file and then combined here.

Overall, we want to go this direction so it's easier to mock the various parts in unit tests.

templates/cluster-template.yaml
```diff
@@ -50,13 +50,13 @@ spec:
     preKubeadmCommands:
       - swapoff -a
   infrastructureTemplate:
```
wongni, line 52:
```suggestion
  machineTemplate:
    infrastructureRef:
```

controllers/cloudstackcluster_controller.go
rejoshed: Removed.

controllers/cloudstackmachine_controller.go
rejoshed: Overall, I believe this to be a false concern. The cloudstack cluster has owner refs all the way down to the machine, and can't be deleted until the machines say they're deleted.

Still, I'm fine moving it back.

I originally needed `csCluster` to specify Account in deletion and fetch calls, but in retrospect I don't think that was necessary, so I moved this back where it was.

rejoshed:
@maxdrib 

Hey, I'm commenting here and adding an @ because can't reply to comments left directly on the PR.

Regarding
> why wouldn't host/port be specified at cluster creation time?
This is just the way we'd designed it. Essentially CAPC grabs an available public IP at runtime and uses that. We can discuss removing the feature if we don't want to support it.

api/v1beta1/cloudstackmachine_webhook.go
rejoshed: Here's an attempted start branch as PR-ed against this one. https://github.com/aws/cluster-api-provider-cloudstack/pull/26/files

Enums in GO really stink, and you can kind of see why in the linked branch.

On top of all the weird extra code to make a type act like an enum, the code still doesn't reject the assignment of a value that's not in the enum's set. So I tried creating a cluster with `Affinity: pr` instead of `pro` and it happily accepted it.

Writing an enum is like writing an entire class for a field value.

pkg/cloud/instance.go
mrog: If the literal is only used once, I'm okay with keeping it as a literal.

api/v1beta1/cloudstackmachine_webhook_test.go
rejoshed: Done.

api/v1beta1/cloudstackcluster_webhook.go
rejoshed: This is just the way it was designed originally. We can change it if we like. It's use case is just grabbing a random IP for the endpoint that's available on the specified network.

api/v1beta1/cloudstackmachine_webhook.go
rejoshed: Cleaned up parens. I think your suggested method is probably a better way to do this, but I don't want to put it through just yet. I think there may be issues with `no` and `""` both being considered unspecified.

controllers/cloudstackmachine_controller.go
rejoshed: I like what you're thinking. That said, I actually want to create scope structs(or maps) that are passed to the cloud package that would contain the log and all other vars. So, I'm leaving the log object encapsulation until/if we get around to that.

We will eventually want to have the log in the cloud package.

controllers/cloudstackmachine_controller.go
rejoshed: It actually just prints what's already printed. So you end up seeing the name of the machine twice.

Apparently DataSecretName is just the name of the machine, which the logger is setup to print from the controller anyway.

controllers/cloudstackmachine_controller.go
rejoshed: Sure, done.

controllers/cloudstackmachine_controller_utils.go
rejoshed: Cool, removed the resolve method and moved all back into the machine controller.

controllers/utils/utils.go
rejoshed: Fixed.

controllers/utils/utils.go
rejoshed: Did both.

pkg/cloud/affinity_groups_test.go
rejoshed: Once we're setup to actually hit a CS instance, then we can simply remove the conditional.

I've in the past also failed if the passing test count wasn't sufficient, so that's a possibility.

Another is labeling these as semi-integ and then running them in a separate job.

Overall, there are a bunch of different ways to handle it once we even get that far.

pkg/cloud/client.go
rejoshed: Just interface composition. : )

pkg/cloud/instance.go
rejoshed: Yup, fixed. Thanks!

pkg/cloud/instance.go
rejoshed: I just added the type of affinity group `pro` or `anti` along with the cluster name.

jweite-amazon (commented):
I'm satisfied that the points I raised were addressed.  Thanks Joshua.  
/approved

Since Max commented significantly, I'd like to get his final approval before we lgtm.

maxdrib (approved):

PR #9, opened 2022-01-19
Compress user_data using gzip to avoid to exceed the hard limit (32KB)

wongni:
*Issue #, if available:*

*Description of changes:*
userData has a limit of 32KB (http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.8/virtual_machines/user-data.html)
When testing with two control planes, three etcd instances, and three worker nodes it failed to create the second control plane because the base64 encoded userData was over 32KB. However, CloudStack supports gzip compression for userData. To avoid too large userData we need to compress the userData using gzip before base64 encoding. 

*Testing performed:*
Manual tests performed. 

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -17,8 +17,6 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
-	"compress/gzip"
 	"context"
 	"encoding/base64"
 	"errors"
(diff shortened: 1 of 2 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -167,7 +169,13 @@ func (r *CloudStackMachineReconciler) reconcile(
 	if !ok {
 		return ctrl.Result{}, errors.New("Bootstrap secret data not ok.")
 	}
-	userData := base64.StdEncoding.EncodeToString(value)
+
+	buf := &bytes.Buffer{}
```
rejoshed, line 173: This all seems reasonable and if you say it works, then I'm good with that.

My only thought is can we turn this into a function called compress data or something similar? I like using methods to abstract these details such that the overall code flow is easier to follow.

rejoshed (changes requested):

controllers/cloudstackmachine_controller.go
wongni: Extracted a function to the cloud package.

PR #15, opened 2022-01-20
Refactoring account/domain validation webhook logic

maxdrib:
*Issue #, if available:*

*Description of changes:*
As I understand, the criteria should be that if account is provided, then domain must be as well. Otherwise, domain is optional regardless of account. This PR attempts to modify the current logic to align with that. 

I encountered this issue after adding the "account" field and trying to spin up a cluster, only specifying domain (as I had been doing prior to this change)

*Testing performed:*
make build && make test

### DIFF
--- a/api/v1alpha3/cloudstackcluster_webhook.go
+++ b/api/v1alpha3/cloudstackcluster_webhook.go
@@ -62,7 +62,7 @@ func (r *CloudStackCluster) ValidateCreate() error {
 		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
-	if (r.Spec.Account == "") != (r.Spec.Domain == "") {
+	if (r.Spec.Account != "") && (r.Spec.Domain == "") {
 		errorList = append(errorList, field.Required(field.NewPath("spec", "account"), "account and domain must be specified together"))
 	}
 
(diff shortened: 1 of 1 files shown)


rejoshed (approved):

api/v1alpha3/cloudstackcluster_webhook.go
```diff
@@ -62,8 +62,8 @@ func (r *CloudStackCluster) ValidateCreate() error {
 		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
-	if (r.Spec.Account == "") != (r.Spec.Domain == "") {
-		errorList = append(errorList, field.Required(field.NewPath("spec", "account"), "account and domain must be specified together"))
+	if (r.Spec.Account != "") && (r.Spec.Domain == "") {
```
jweite-amazon, line 65: Why the change?  The original construct is surprising, but effectively saying that if account == "" is true then domain must == "" must also be true, and if account == "" is false then domain must == "" must also be false.

The changed logic doesn't detect domain provided but account blank.

jweite-amazon: I just read the PR description.  I hadn't understood it quite they way you described it.  Let me think on this a minute...

jweite-amazon: I stand corrected.  The API Ref states that domain must be provided if account is provided, but nothing about the other way around.

PR #19, opened 2022-01-25
Feature/health check v ms

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Added templates for adding control plane and worker machine health checks.  These are intentionally segregated from the provided clusterTemplates, for use by deployers as they see fit.

*Testing performed:*
Manual testing of unhealthy detection and remediation by stopping ACS VMs and monitoring the remediation process.

### DIFF
--- a/templates/controlPlaneMachineHealthCheck.yaml
+++ b/templates/controlPlaneMachineHealthCheck.yaml
@@ -1,0 +1,31 @@
+# This custom resource can be optionally be defined for a cluster, enabling CAPI Machine Health Checking
+# for its control plane nodes.  See https://cluster-api.sigs.k8s.io/tasks/healthcheck.html
+#
+# This has intentionally been segregated from the cluster-templates, as it introduces a few complexities
+# surrounding CNIs:
+#  - The CNI must be deployed and the cluster nodes become ready within the nodeStartTimeout (or else
+#    the MachineHealthCheck remediation processes will begin terminating the unready worker nodes).
+#  - Certain CNIs have been observed to hang the MachineHealthCheck remediation processes's attempts
+#    to delete the failed node (inability to drain).
+# As such, the deployment of this component is left to the discretion of the cluster deployer.
+# If deployed independently of cluster-template, be sure to replace the placeholders in the below
+# before applying it to your management cluster.
+---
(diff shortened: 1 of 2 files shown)


rejoshed (approved):
Could see putting this in one file, but I'm happy either way. Looks good!

rejoshed:
Can ltgm if you'd like, or feel free to add the label yourself. : )

templates/controlPlaneMachineHealthCheck.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: ${CLUSTER_NAME}-kcp-unhealthy-2m
+spec:
+  clusterName: ${CLUSTER_NAME}
+  maxUnhealthy: 100%
+#  nodeStartupTimeout: 10m
+  selector:
+    matchLabels:
+      cluster.x-k8s.io/control-plane: ""
+  unhealthyConditions:
```
rejoshed, line 25: Kidna weird that we're using the negative here. Would positive health checks be more complete? Is that possible?

jweite-amazon: That's just how the CAPI MachineHealthCheck CRD is defined: https://github.com/kubernetes-sigs/cluster-api/blob/58905fb5140a7844c3e24aa70d255fb8f9d0c491/api/v1alpha3/machinehealthcheck_types.go#L41

maxdrib:
Why not include these objects by default, and segregate them instead?

jweite-amazon:
@maxdrib, I don't follow how to include-yet-segregate.  What did you have in mind?

maxdrib:
> @maxdrib, I don't follow how to include-yet-segregate. What did you have in mind?

I just mean why are we segregating them? Why not put them in the default templates? Are there any downsides of including this as a default object installed on a CAPC cluster?

jweite-amazon:
@maxdrib, comments in the file elaborate on why I elected to segregate.  Most significantly is that "certain CNIs" (I didn't want to call out Cilium by name) interfere with the deletion of a failed worker node.  I don't want to cause that to happen for folks who may not fully understand what's the mechanism.  Instead I prefer to let them explicitly elect to try health checking and then discover whether it's compatible with their CNI or not.

PR #20, opened 2022-01-26
Feature/v1beta1

rejoshed:
*Issue #* Bump to v1beta1 -- can't get tickets to load atm, so can't get number.

*Description of changes:*
Upgraded to support CAPI v1beta1 resources. See comments in `Files Changed` tab for further descriptions.

*Testing performed:*
Created a cluster in a shared network.
Deleted the cluster.
Created a cluster in an isolated network.
Scaled from 1:1 nodes to 2:3 back to 1:1.
Deleted the cluster.


config/rbac/service_account.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: controller-manager
```
maxdrib, line 4: why not controller manager? I believe other CAPI providers have separate docker images for a controller vs manager. In our case, they're combined. Shouldn't the name reflect that?

maxdrib (commented):
Given the CAPI version upgrade, should we do a minor version bump to 0.4, with an updated entry in metadata.yaml? That should make it easier to work with clusterctl as well. Otherwise  code LGTM, I'd just be curious to see how it was tested.

config/rbac/service_account.yaml
rejoshed: I was following along with the Cluster Api book's upgrade destructions. They used the name manager, and I wasn't sure whether it was relevant or not.

This change wasn't intended for the final PR unless it was necessary.

I just swapped it back as it wasn't.

rejoshed:
@maxdrib 

> Otherwise code LGTM, I'd just be curious to see how it was tested.

It wasn't! Not at the time you reviewed anyway. It's still draft atm, but I'll get a description and write about some of my testing in the final PR. : )

Currently I've created and deleted a cluster. I'll probably try scaling up and down before the final PR.

Makefile
```diff
@@ -134,7 +134,7 @@ undeploy: bin/kustomize ## Undeploy controller from the K8s cluster specified in
 .PHONY: binaries
 binaries: bin/controller-gen bin/kustomize bin/ginkgo bin/golangci-lint bin/mockgen bin/kubectl ## Locally install all needed bins.
 bin/controller-gen: ## Install controller-gen to bin.
-	GOBIN=$(PROJECT_DIR)/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.9
+	GOBIN=$(PROJECT_DIR)/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0
```
rejoshed, line 137: Latest release is 0.80.

Makefile
```diff
@@ -181,19 +181,12 @@ tilt-up: cluster-api kind-cluster cluster-api/tilt-settings.json manifests cloud
 	export CLOUDSTACK_B64ENCODED_SECRET=$$(base64 -i cloud-config) && cd cluster-api && tilt up
 
 .PHONY: kind-cluster
-kind-cluster: cluster-api cluster-api/hack/kind-install-for-capd.sh # Create a kind cluster with a local Docker repository.
+kind-cluster: cluster-api # Create a kind cluster with a local Docker repository.
```
rejoshed, line 184: Since we're now on v1.+, we no longer have to checkout the kind install file as it's present by default.

api/v1beta1/cloudstackcluster_webhook.go
```diff
@@ -37,7 +37,7 @@ func (r *CloudStackCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-//+kubebuilder:webhook:path=/mutate-infrastructure-cluster-x-k8s-io-v1alpha3-cloudstackcluster,mutating=true,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=cloudstackclusters,verbs=create;update,versions=v1alpha3,name=mcloudstackcluster.kb.io
+//+kubebuilder:webhook:path=/mutate-infrastructure-cluster-x-k8s-io-v1beta1-cloudstackcluster,mutating=true,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=cloudstackclusters,verbs=create;update,versions=v1beta1,name=mcloudstackcluster.kb.io,admissionReviewVersions=v1beta1
```
rejoshed, line 40: I did not see admissionReviewVersion listed in the CAPI book's upgrade destructions, so it took me a bit to realize it was needed. Grr...

api/v1beta1/webhook_suite_test.go
```diff
@@ -56,16 +54,14 @@ func TestAPIs(t *testing.T) {
...
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
-
 	ctx, cancel = context.WithCancel(context.TODO())
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "config", "crd", "bases")},
 		ErrorIfCRDPathMissing: false,
 		WebhookInstallOptions: envtest.WebhookInstallOptions{
-			DirectoryPaths: []string{filepath.Join("..", "..", "config", "webhook")},
```
rejoshed, line 68: API name change apparently.

api/v1beta1/webhook_suite_test.go
```diff
@@ -35,8 +35,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
rejoshed, line 39: Consolidated our logging to use only klog. This seems to be the default across providers.

config/dev/insecure_manager.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+  labels:
```
rejoshed, line 6: I moved this patch here since we're using the newer Tiltfile now. Unfortunately, both old and new Tiltfiles don't really provide an option to point to a different entry point for kustomize. I will do some more digging eventually, but for now we'll still need to manually add and remove the nonRoot param when using Tilt.

config/webhook/manifests.yaml
```diff
@@ -1,127 +1,111 @@
-
 ---
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: MutatingWebhookConfiguration
 metadata:
```
rejoshed, line 4: The file changes were generated by controller-gen.

controllers/cloudstackcluster_controller.go
```diff
@@ -63,8 +63,7 @@ type CloudStackClusterReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
-func (r *CloudStackClusterReconciler) Reconcile(req ctrl.Request) (retRes ctrl.Result, retErr error) {
-	ctx := context.Background()
+func (r *CloudStackClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (retRes ctrl.Result, retErr error) {
```
rejoshed, line 66: Interface method signature is different for v1beta1. Context is passed instead.

controllers/cloudstackcluster_controller.go
```diff
@@ -168,28 +167,23 @@ func (r *CloudStackClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		return err
 	}
-
-	// Add a watch on CAPI Cluster objects for unpause and ready events.
-	return controller.Watch(
+	return controller.Watch( // Add a watch on CAPI Cluster objects for unpause and ready events.
 		&source.Kind{Type: &capiv1.Cluster{}},
-		&handler.EnqueueRequestsFromMapFunc{
-			ToRequests: util.ClusterToInfrastructureMapFunc(infrav1.GroupVersion.WithKind("CloudStackCluster"))},
+		handler.EnqueueRequestsFromMapFunc(
```
rejoshed, line 176: EnqueuRequestsFromMapFunc is now a method and not a type.

main.go
```diff
@@ -51,33 +52,70 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
-func main() {
+type managerOpts struct {
```
rejoshed, line 55: I really didn't like that all the options passed via args were just individual vars.

In use these vars did not indicate that they were input from the user at startup.

Now instead of cloudConfigFile we have opts.CloudConfigFile. This also helps to remove the opt vars from main, so for clarity alone, this is a win!

main.go
```diff
@@ -17,24 +17,25 @@ limitations under the License.
...
-	"github.com/aws/cluster-api-provider-cloudstack/pkg/cloud"
+	"fmt"
 	"os"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	"strings"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	// to ensure that exec-entrypoint and run can make use of them.
+	"github.com/aws/cluster-api-provider-cloudstack/pkg/cloud"
+	flag "github.com/spf13/pflag"
```
rejoshed, line 25: pflag is a 'drop in' replacement for the builtin flag that supports posix compatible flags.

So --metrics-addr instead of -metrics-addr.

main.go
```diff
@@ -51,33 +52,70 @@ func init() {
...
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
-	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+func setFlags() *managerOpts {
+	opts := &managerOpts{}
+	flag.StringVar(
+		&opts.CloudConfigFile,
+		"cloud-config-file",
+		"/config/cloud-config",
+		"Overrides the default path to the cloud-config file that contains the CloudStack credentials.")
+	flag.StringVar(
```
rejoshed, line 72: There are a few changes to input params as the upgrade desctructions required renaming and or adding and or change of default values.

Nothing to important though.

config/dev/insecure_manager.yaml
jweite-amazon: I don't see this dev "configuration module" getting pulled in anywhere.  Is this just for posterity?

metadata.yaml
```diff
@@ -14,4 +14,7 @@ releaseSeries:
     contract: v1alpha3
   - major: 0
     minor: 3
-    contract: v1alpha3
\ No newline at end of file
+    contract: v1alpha3
+  - major: 0
```
jweite-amazon, line 18: Should we still be including the v1alpha3 refs if we no longer support it?

jweite-amazon:
The one surprise for me here was that we aren't offering v1alpha3 as an alternative supported API version anymore.  I had imagined you were adding v1beta1 as a new hub, with conversion to/from v1alpha3.

I'm on the fence about whether it's needed / worth the effort, given how cross-version applications seem to be blocked by other CAPI-isms like clusterctl.

config/dev/insecure_manager.yaml
rejoshed: It's intended to be an overlay. If you use `kustomize config/dev` instead of `config/default` nonRoot will be set to false. Eventually we'll want to have the tilt point here.

So, yeah kind for posterity. Basically we should figure out how to make tilt use it, but I haven't yet found a way outside of modifying the Tiltfile directly.

metadata.yaml
rejoshed: I'm not really sure. I just made this match other providers in the way they set this.

jweite-amazon: Gotcha.  They offer backward-compatible v1alpha3/4 APIs, though.  If/since we're no longer offering those API versions, I'd rather we just make like they never existed.

rejoshed: Okay, I'll just delete the v1alpha3 stuffs with early versionings.

go.sum
```diff
@@ -1,331 +1,469 @@
...
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
-github.com/creack/pty v1.1.7/go.mod h1:lj5s0c3V2DBrqTV7llrYr5NG6My20zk30Fl46Y7DoTY=
-github.com/davecgh/go-spew v0.0.0-20151105211317-5215b55f46b2/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
+github.com/creack/pty v1.1.11/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/daviddengcn/go-colortext v0.0.0-20160507010035-511bcaf42ccd/go.mod h1:dv4zxwHi5C/8AeI+4gX4dCWOIvNi7I6JCSX0HvlKPgE=
+github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
```
maxdrib: this library is bad according to dependabot. Let's keep the replace in the go.mod

metadata.yaml
maxdrib: to be clear, this commit itself won't serve as a version bump/release, right? We'll keep developing against it before we cut a final tag/release?

maxdrib:
Can we confirm with the customer that not supporting v1alpha3 is acceptable before we merge this change?

metadata.yaml
rejoshed: Yes, for sure. : )

PR #22, opened 2022-01-31
Exposing controller metrics on port 8080 on all IPs, a la CAPV.

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Exposing CAPC Controller Metrics port for direct access, a la CAPV (no kube-proxy, no Prometheus Operator ServiceMonitor CRD).

*Testing performed:*

### DIFF
--- a/config/default/manager_prometheus_metrics_patch.yaml
+++ b/config/default/manager_prometheus_metrics_patch.yaml
@@ -1,0 +1,19 @@
+# This patch enables Prometheus scraping for the manager pod.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: 'true'
+    spec:
+      containers:
+      # Expose the prometheus metrics on default port
+      - name: manager
+        ports:
+        - containerPort: 8080
+          name: metrics
+          protocol: TCP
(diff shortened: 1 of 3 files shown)


config/default/manager_prometheus_metrics_patch.yaml
```diff
@@ -0,0 +1,19 @@
+# This patch enables Prometheus scraping for the manager pod.
```
jweite-amazon, line 1: From CAPV, verbatim.

jweite-amazon:
The default --metrics-addr of localhost:8080 makes the metrics port only available to sidecar containers in the same pod as the controller (typically kube-rbac-proxy).  I originally opened this port up (by binding to all IP addresses), but based on CAPV code reverted this to keep the port more secure by default.  (Info leakage through metrics can be a risk).  Deployers wishing to expose the metrics port more broadly can elect to either use kube-rbac-proxy or expose it on all IPs (with --metrics-addr=:8080) in an override version of infrastructure-components.yaml.

rejoshed:
Well, I ran through the tests with no issue 5 times in a row, but now it's failed twice in a row?

My guess is that the Prow server is struggling with load? Unfortunately, since we have no visibility I can't check into that.

The test isn't particularly important since we do manual testing and this test is _the_ most basic check that the controller is working. We could just turn it off until we can do more investigation.

jweite-amazon:
> The test isn't particularly important since we do manual testing and this test is _the_ most basic check that the controller is working. We could just turn it off until we can do more investigation.

Agreed that the false-failure noise is distracting, but I kinda prefer to leave it in place as a reminder, both to fix the Prow job, and to manually unit test the PR.

PR #25, opened 2022-02-08
Minor code cleanup

mrog:
*Issue #, if available:* CTECH-135

*Description of changes:* Minor code cleanup, including use of constants and fixing typos

*Testing performed:* Code built

### DIFF
--- a/api/v1beta1/cloudstackcluster_webhook_test.go
+++ b/api/v1beta1/cloudstackcluster_webhook_test.go
@@ -20,20 +20,19 @@ import (
 	"context"
 
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 var _ = Describe("CloudStackCluster webhooks", func() {
 	const (
-		apiVersion       = "infrastructure.cluster.x-k8s.io/v1beta1"
-		clusterKind      = "CloudStackCluster"
-		clusterName      = "test-cluster"
-		clusterNamespace = "default"
-		identityName     = "IdentitySecret"
-		zone             = "Zone"
-		network          = "Network"
+		apiVersion         = "infrastructure.cluster.x-k8s.io/v1beta1"
+		clusterKind        = "CloudStackCluster"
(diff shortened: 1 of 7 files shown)


api/v1beta1/cloudstackcluster_webhook_test.go
```diff
@@ -20,30 +20,39 @@ import (
 	"context"
 
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var _ = Describe("CloudStackCluster webhooks", func() {
+	const (
```
rejoshed, line 27: I'm okay with it, but I really don't think consts are necessary like this in test files.

pkg/cloud/instance.go
```diff
@@ -163,9 +163,9 @@ func (c *client) GetOrCreateVMInstance(
 	setIfNotEmpty(csCluster.Spec.Account, p.SetAccount)
 	setIfNotEmpty(csCluster.Status.DomainID, p.SetDomainid)
 
-	// If this VM instance is a control plane, consider setting it's IP.
```
rejoshed, line 166: Nice.

rejoshed:
TBH, not sure I like the consts in the testing files, but not sure I care that much.

rejoshed (approved):

api/v1beta1/cloudstackcluster_webhook_test.go
mrog: They'd certainly be more important in production code, but I've experienced a fair number of test failures (or even worse, tests that don't fail when they should) due to mismatched literals.  If the same literal is used more than once, or is likely to be used again in the future, I like to use a constant or enum.

pkg/cloud/network.go
```diff
@@ -31,6 +31,7 @@ const (
 	K8sDefaultAPIPort   = 6443
 	NetworkTypeIsolated = "Isolated"
 	NetworkTypeShared   = "Shared"
+	tcp                 = "tcp"
```
wongni, line 34:
```suggestion
	NetworkProtocolTCP                 = "tcp"
```
To be consistent with other constants.

mrog: Good idea.  Thanks!

PR #28, opened 2022-02-10
feature/tag-network

rejoshed:
*Issue #, if available:*

*Description of changes:*

*Testing performed:*


pkg/cloud/affinity_groups_test.go
```diff
@@ -49,8 +49,9 @@ var _ = Describe("AffinityGroup Unit Tests", func() {
 		fakeAG = &cloud.AffinityGroup{
 			Name: "FakeAffinityGroup",
 			Type: cloud.AffinityGroupType}
-		cluster = &infrav1.CloudStackCluster{Spec: infrav1.CloudStackClusterSpec{
-			Zone: "Zone1", Network: "SharedGuestNet1"}}
+		cluster = &infrav1.CloudStackCluster{
+			Spec:       infrav1.CloudStackClusterSpec{Zone: "Zone1", Network: "SharedGuestNet1"},
```
rejoshed, line 53: I'm fine with this, but you should be able to skip the import and instead do `cluster.ObjectMeta.SetUID("0")`

pkg/cloud/network.go
```diff
@@ -58,9 +58,38 @@ func (c *client) ResolveNetwork(csCluster *infrav1.CloudStackCluster) (retErr er
...
 }
 
+func generateNetworkTagName(csCluster *infrav1.CloudStackCluster) string {
+	return clusterTagNamePrefix + string(csCluster.UID)
+}
+
 func (c *client) GetOrCreateNetwork(csCluster *infrav1.CloudStackCluster) (retErr error) {
-	if retErr = c.ResolveNetwork(csCluster); retErr == nil { // Found network.
+	// Tag the network so we can clean it up later
+	tagNetwork := func(csCluster *infrav1.CloudStackCluster, addCreatedByTag bool) error {
```
rejoshed, line 67: Not entirely sure why we're defining the func on the fly. Was there some closure goodness here that was needed?

pkg/cloud/network.go
```diff
@@ -86,6 +115,32 @@ func (c *client) GetOrCreateNetwork(csCluster *infrav1.CloudStackCluster) (retEr
 	csCluster.Status.NetworkID = resp.Id
 	csCluster.Status.NetworkType = resp.Type
 
+	return tagNetwork(csCluster, true)
+}
+
+func (c *client) UntagAndDestroyNetwork(csCluster *infrav1.CloudStackCluster) (retError error) {
```
rejoshed, line 121: I think I'd prefer these to be two separate methods. An untag and then a deleteIfNotInUse or something like that.

pkg/cloud/network_test.go
```diff
@@ -108,6 +122,10 @@ var _ = Describe("Network", func() {
 			ns.EXPECT().NewCreateNetworkParams(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
 				Return(&cloudstack.CreateNetworkParams{})
 			ns.EXPECT().CreateNetwork(gomock.Any()).Return(&cloudstack.CreateNetworkResponse{Id: netId}, nil)
+			rs.EXPECT().NewListTagsParams().Return(&cloudstack.ListTagsParams{})
+			rs.EXPECT().ListTags(gomock.Any()).Return(&cloudstack.ListTagsResponse{}, nil)
+			rs.EXPECT().NewCreateTagsParams(gomock.Any(), gomock.Any(), gomock.Any()).Return(&cloudstack.CreateTagsParams{})
```
rejoshed, line 127: Could use the param matchers I built here so that we're checking the params are reasonable. No biggie though.

pkg/cloud/network_test.go
```diff
@@ -87,13 +93,21 @@ var _ = Describe("Network", func() {
 		It("does not call to create a new network via GetOrCreateNetwork", func() {
 			ns.EXPECT().GetNetworkID(fakeNetName).Return(fakeNetId, 1, nil)
 			ns.EXPECT().GetNetworkByID(fakeNetId).Return(&cloudstack.Network{Type: isolatedNetworkType}, 1, nil)
+			rs.EXPECT().NewListTagsParams().Return(&cloudstack.ListTagsParams{})
```
rejoshed, line 96: These expects are not an in-order thing. So, since they're duplicated for each of these tests, we could move the new expects to a `BeforeEach` function.

pkg/cloud/tags.go
```diff
@@ -0,0 +1,63 @@
...
+
+const (
+	clusterTagNamePrefix = "CAPC_cluster_"
+	createdByCapcTagName = "created_by_CAPC"
+	resourceTypeNetwork  = "network"
+)
+
+// TagNetwork adds tags to a network by network id.
+func (c *client) AddNetworkTags(networkId string, tags map[string]string) error {
+	// https://cloudstack.apache.org/api/apidocs-4.16/apis/createTags.html
```
rejoshed, line 33: Let's go ahead and remove the links to the api documentation here. I just put them in for reference when you were getting started. : )

pkg/cloud/tags_test.go
```diff
@@ -0,0 +1,88 @@
...
+			networkId string
+			testTags  map[string]string
+		)
+
+		BeforeEach(func() {
+			if connectionErr != nil { // Only do these tests if an actual ACS instance is available via cloud-config.
+				Skip("Could not connect to ACS instance.")
+			}
+
+			if err := client.GetOrCreateNetwork(cluster); err != nil {
```
rejoshed, line 61: I think in these circumstances we should just create one and delete it in the after block.

I won't gate the PR for it, but it's a nice to have.

Once we get our tests running proper in the pipeline, these tests won't actually be skipped, but will hit our EC2 instance.

pkg/cloud/network.go
mrog: Just because it's not needed outside this scope, and it started out as a really simple function.  But now that's is grown, it makes more sense to define it outside the function.  I'll move it.

pkg/cloud/affinity_groups_test.go
mrog: That sounds reasonable.  I'll do it.

pkg/cloud/network.go
mrog: That would make the intent clearer.  Good idea!

pkg/cloud/network_test.go
mrog: Good plan.  I'll do it.

pkg/cloud/network_test.go
mrog: Argument matches on mock stubs can make tests difficult to debug.  If I were to check the args, I'd prefer to do it in a verification stub at the end of the test.  But it doesn't look like gomock provides that ability.  (FWIW, it was requested 3 years ago.  https://github.com/golang/mock/issues/262)

I can throw a few checks in there, anyway, for data that's shared by all the tests.  (I'm moving these lines into a BeforeEach.

pkg/cloud/cluster.go
```diff
@@ -92,3 +93,10 @@ func (c *client) GetOrCreateCluster(csCluster *infrav1.CloudStackCluster) (retEr
 	csCluster.Status.Ready = true
 	return nil
 }
+
+func (c *client) DisposeClusterResources(csCluster *infrav1.CloudStackCluster) (retError error) {
+	if err := c.RemoveClusterTagFromNetwork(csCluster); err != nil {
+		return err
+	}
+	return c.DeleteNetworkIfNotInUse(csCluster)
```
rejoshed, line 101: Nice.

PR #30, opened 2022-02-15
Feature/ssl opt

rejoshed:
*Issue #, if available:* TBD

*Description of changes:*
Added the option to set the verify-ssl boolean found on the ACS client. See [here](https://pkg.go.dev/github.com/xanzy/go-cloudstack@v2.4.1+incompatible/cloudstack#NewAsyncClient)

*Note:* This does not enforce the use of an https based protocol. So for the verification to matter, the ACS instance needs to be served at an https endpoint.

The new input config can be written as:
``` ini
[Global]
api-key    = [apiKey] 
secret-key = [secretKey]
api-url    = [https://someAddr:8080/client/api]
verify-ssl = [false|true]
```

And `verify-ssl` is optional, but defaults to true.

*Testing performed:*
Connected via the manager insecurely via http and attempted securely via https protocol. I don't yet have an https version of ACS to test against though.


.gitignore
```diff
@@ -7,6 +7,9 @@
 bin
 testbin/*
 
+pkg/mocks/*
```
rejoshed, line 10: I also removed the generated mocks from the package.

pkg/cloud/client.go
```diff
@@ -45,37 +46,37 @@ type client struct {
 	// csA *cloudstack.CloudStackClient
 }
 
+// cloud-config ini structure.
+type config struct {
```
rejoshed, line 50: Turns out there's a perfectly natural way to use the `ini` lib that I wasn't using. Derp.

pkg/cloud/client.go
```diff
@@ -45,37 +46,37 @@ type client struct {
...
+	c.cs = cloudstack.NewAsyncClient(cfg.ApiUrl, cfg.ApiKey, cfg.SecretKey, cfg.VerifySSL)
 	_, err = c.cs.Zone.ListZones(c.cs.Zone.NewListZonesParams())
 	if err != nil && strings.Contains(err.Error(), "i/o timeout") {
 		return c, errors.Wrap(err, "Timeout while checking CloudStack API Client connectivity.")
 	}
 	return c, errors.Wrap(err, "Error encountered while checking CloudStack API Client connectivity.")
 }
 
-// CloudStack API config reader.
-func readAPIConfig(cc_path string) (string, string, string, error) {
```
rejoshed, line 67: Just not needed when using the `ini` lib properly.

Makefile
```diff
@@ -170,7 +170,7 @@ test: generate-mocks lint generate-deepcopy bin/ginkgo bin/kubectl bin/kube-apis
 		./hack/testing_ginkgo_recover_statements.sh --remove; exit $$EXIT_STATUS
 	
 .PHONY: generate-mocks
-generate-mocks: bin/mockgen $(shell find ./pkg/mocks -type f -name "mock*.go") ## Generate mocks needed for testing. Primarily mocks of the cloud package.
+generate-mocks: bin/mockgen pkg/mocks/mock_client.go $(shell find ./pkg/mocks -type f -name "mock*.go") ## Generate mocks needed for testing. Primarily mocks of the cloud package.
```
wongni, line 173: pkg/mocks/mock_client.go should be returned from $(shell find ./pkg/mocks -type f -name "mock*.go")
Is this additional dependency really required?

pkg/cloud/client_test.go
```diff
@@ -0,0 +1,59 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloud_test
+
+// import (
```
wongni, line 19: Why don't you remove this file as this file is technically empty?

rejoshed: Whoops. Didn't mean for it to be commented out like that.

Was trying to figure out why the suite was hanging. It's fixed now.

Makefile
rejoshed: It's not found from that if the file doesn't exist to begin with.

pkg/cloud/client.go
jweite-amazon: Good to leverage this.
Looking down the road toward multi-endpoint, I envisioned having a section-per-endpoint, with a user-provided logical name for each section (by which that endpoint would be referred to).  Does the ini lib provide any means for enumerating the sections in an ini file if they're not known a-priori?

wongni (approved):

PR #31, opened 2022-02-16
feature/failure domains

rejoshed:
*Issue #, if available:* TBD

*Note*
This is a biggie. I'm probably going to continue minor cleanups during the PR process. At this point it works, but I want to fixup things like the add_address.sh script an other small details.

*Description of changes:*
Adds multi-zone support via failure domains.

Adds dummies module that provides test dummies. Simplifies testing quite a bit.

Some small changes to tilt config and addition of further config/dev entrypoint kustomization patches.

Removal of generated deepcopy.

*Testing performed:*
image

Can now create a cluster across Zones. Also tested isolated network creation in both the case of creating network and not.


api/v1beta1/cloudstackcluster_types.go
```diff
@@ -39,13 +39,35 @@ type CloudStackIdentityReference struct {
 	Name string `json:"name"`
 }
 
+type Network struct {
```
jweite-amazon, line 42: I'm surprised to see some of the attributes being used for the spec.  I'd imagined ZoneId, Zone.NetworkId and Zone.NetworkType to be status-only attributes.

api/v1beta1/cloudstackcluster_webhook_test.go
```diff
@@ -14,198 +14,85 @@ See the License for the specific language governing permissions and
...
-						Name: identitySecretName,
-					},
-					Network: network,
-				},
-			}
-			Expect(k8sClient.Create(ctx, cloudStackCluster).Error()).Should(MatchRegexp("admission webhook.*denied the request.*Required value\\: Zone"))
+		It("Should reject a CloudStackCluster with missing Zones.Network attribute", func() {
+			dummies.CSCluster.Spec.Zones = []infrav1.Zone{{Name: "ZoneWNoNetwork"}}
+			Ω(k8sClient.Create(ctx, dummies.CSCluster)).Should(
+				MatchError(MatchRegexp(requiredRegex, "each Zone requires a Network specification")))
```
rejoshed, line 48: Moved all of these tests to use
`Ω(<some func>).Should(MatchError(MatchRegexp(<regex>)))`

Instead of
`Ω(<some func>.Error()).Should(MatchRegexp(<regex>))`.

The old style would panic if some func returned nil.

api/v1beta1/cloudstackmachine_types.go
```diff
@@ -52,14 +52,14 @@ type CloudStackMachineSpec struct {
 
 	// Optional affinitygroupids for deployVirtualMachine
 	// +optional
-	AffinityGroupIds []string `json:"affinitygroupids,omitempty"`
+	AffinityGroupIDs []string `json:"affinitygroupids,omitempty"`
```
rejoshed, line 55: Linter prefers ID almost everywhere apparently.

tilt-provider.json
```diff
@@ -1,7 +1,8 @@
 {
     "name": "cloudstack",
-    "label": "CAPC",
     "config": {
+        "manager_name": "capc-controller-manager",
```
rejoshed, line 4: This change just fixes the labeling in tilt.

api/v1beta1/cloudstackcluster_types.go
maxdrib: +1 @rejoshed do all these have to passed in from eks-a? Or just some of them? I thought the input was ambiguous and capc would itself determine if the input was a name or ID

jweite-amazon:
_make docker build_ needs a dependency on _generate-deepcopy_.  Otherwise it fails after _make clean._

api/v1beta1/cloudstackcluster_types.go
rejoshed: Can pass either or, but do need to specify the correct one. So, id needs to be the ACS ID and name has to match the name.

api/v1beta1/cloudstackcluster_webhook.go
```diff
@@ -43,7 +44,7 @@ var _ webhook.Defaulter = &CloudStackCluster{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *CloudStackCluster) Default() {
-	cloudstackclusterlog.Info("default", "name", r.Name)
+	cloudstackclusterlog.V(1).Info("entered api default setting webhook", "api resource name", r.Name)
```
rejoshed, line 47: These log lines seemed rather noisy. Bumped to V(1)

api/v1beta1/cloudstackcluster_webhook.go
```diff
@@ -63,23 +64,31 @@ func (r *CloudStackCluster) ValidateCreate() error {
 	}
 
 	if (r.Spec.Account != "") && (r.Spec.Domain == "") {
-		errorList = append(errorList, field.Required(field.NewPath("spec", "account"), "specifying account requires additionally specifying domain"))
+		errorList = append(errorList, field.Required(
+			field.NewPath("spec", "account"), "specifying account requires additionally specifying domain"))
```
rejoshed, line 68: I'm gradually trying to move us to 120 char limit. Wrapped line here and lots of other places.

Makefile
```diff
@@ -74,16 +74,12 @@ api/%/zz_generated.deepcopy.go: bin/controller-gen $(DEEPCOPY_GEN_INPUTS)
 
 MANAGER_BIN_INPUTS=$(shell find ./controllers ./api ./pkg -name "*mock*" -prune -o -name "*test*" -prune -o -type f -print) main.go go.mod go.sum
 .PHONY: build
-build: binaries generate-mocks generate-deepcopy manifests release-manifests ## Build manager binary.
+build: binaries generate-deepcopy lint manifests release-manifests ## Build manager binary.
```
rejoshed, line 77: @jweite-amazon Fixup makefile for generation of deepcopy files.

Actually pretty happy that I removed the generated files and thus created this bug. Means that the deepcopy stuff wasn't properly set in the Makefile, but that we were just typically getting lucky that the deepcopy stuff was present and not broken.

api/v1beta1/cloudstackcluster_webhook_test.go
```diff
@@ -14,198 +14,85 @@ See the License for the specific language governing permissions and
...
 
-				cloudStackCluster.DeepCopyInto(cloudStackClusterUpdate)
-				cloudStackClusterUpdate.Spec.IdentityRef.Name = configMapName
-				Expect(k8sClient.Update(ctx, cloudStackClusterUpdate).Error()).Should(MatchRegexp(forbiddenRegex, "identityRef\\.Name"))
-			})
+		It("Should reject updates to CloudStackCluster Zones", func() {
+			dummies.CSCluster.Spec.Zones = []infrav1.Zone{dummies.Zone1}
+			Ω(k8sClient.Update(ctx, dummies.CSCluster)).Should(MatchError(MatchRegexp(forbiddenRegex, "Zones and sub")))
+		})
+		It("Should reject updates to CloudStackCluster Zones", func() {
```
jweite-amazon, line 72: Copy/paste from above?  It should reject updates to CloudStackCluster Zone's Networks.

api/v1beta1/cloudstackcluster_webhook_test.go
```diff
@@ -14,198 +14,85 @@ See the License for the specific language governing permissions and
...
+			Ω(k8sClient.Update(ctx, dummies.CSCluster)).
+				Should(MatchError(MatchRegexp(forbiddenRegex, "controlplaneendpoint\\.port")))
+		})
+		It("Should reject updates to the CloudStackCluster identity reference kind", func() {
+			dummies.CSCluster.Spec.IdentityRef.Kind = "ConfigMap"
+			Ω(k8sClient.Update(ctx, dummies.CSCluster)).
+				Should(MatchError(MatchRegexp(forbiddenRegex, "identityref\\.kind")))
+		})
+		It("Should reject updates to the CloudStackCluster identity reference name", func() {
+			dummies.CSCluster.Spec.IdentityRef.Name = "ConfigMap"
```
jweite-amazon, line 93: Setting the Name to "ConfigMap" feels wrong, given that it's a type name, even though it works.  I'd rather the name be "NewName" or something like that.

controllers/cloudstackmachine_controller.go
```diff
@@ -153,41 +156,55 @@ func (r *CloudStackMachineReconciler) reconcile(
...
+	if util.IsControlPlaneMachine(capiMachine) { // Use failure domain zone.
+		csMachine.Status.ZoneID = *capiMachine.Spec.FailureDomain
+	} else { // Random zone.
+		zones := make([]string, len(csCluster.Status.Zones))
+		zidx := 0
+		for zoneID := range csCluster.Status.Zones {
+			zones[zidx] = zoneID
+			zidx++
+		}
+		rand.Seed(time.Now().Unix())
```
jweite-amazon, line 176: I believe re-seeding the rand every invocation thwarts its randomness.  And is a cost that should only be borne once.  I think this should be called in main.go.

api/v1beta1/cloudstackcluster_webhook_test.go
rejoshed: Ah, I should update the it statement.

In the first one I change the Zones entirely and in the second I change the network within a zone.

I'm working through a merge against main atm, but I'll update the description shortly.

api/v1beta1/cloudstackcluster_webhook_test.go
rejoshed: I just did what the previous test was doing, but I can change it.

Basically it should always be "Secret" for us.

Do you want me to change it to "NewName"?

test/dummies/vars.go
```diff
@@ -0,0 +1,254 @@
...
+	Tag1Key = "test_tag1"
+	Tag1Val = "arbitrary_value1"
+	Tag2Key = "test_tag2"
+	Tag2Val = "arbitrary_value2"
+	Tag1 = map[string]string{Tag2Key: Tag2Val}
+	Tag2 = map[string]string{Tag2Key: Tag2Val}
+	Tags = map[string]string{Tag1Key: Tag1Val, Tag2Key: Tag2Val}
+}
+
+// SetDummyClusterSpecVars resets the values in each of the exported CloudStackMachines related dummy variables.
```
rejoshed, line 95: Note to self: Update comments for functions in this file. Many don't match the function name.

controllers/cloudstackmachine_controller.go
rejoshed: I think it does the opposite since the time it's seeded is random too, but I can move it to main if you'd like?

api/v1beta1/cloudstackcluster_webhook_test.go
jweite-amazon: Yes.  Otherwise it looks like a mistake to my eyes until I study it more carefully.  I'd rather other readers don't get distracted by it.

controllers/cloudstackmachine_controller.go
jweite-amazon: Yes, please.  I'd rather put my trust in the randomness of the algorithm.

.gitignore
```diff
@@ -6,6 +6,9 @@
 *.dylib
 bin
 testbin/*
+api/*/zz*generated.deepcopy.go
```
maxdrib, line 9: can we avoid removing this for now? I'm not sure what kind of magic will be required for eks-a to work without it. Currently, this is all we're running before expecting the repo to be functional: https://github.com/maxdrib/eks-anywhere-build-tooling/blob/main/projects/aws/cluster-api-provider-cloudstack/Makefile#L22-L24

Makefile
maxdrib, line 77: why are we not generating mocks any more as part of this? Again, eks-a might not like this very much. We need to be careful about what changes are being made to the build scripts since they have a dependency

rejoshed: > why are we not generating mocks any more as part of this?
We are.

On line 169 `generate-mocks: bin/mockgen generate-deepcopy pkg/mocks/mock_client.go ` generate mocks generates deepcopy as a prerequisite. You shouldn't have to change the EKS-A command. 

Please let me know if you do.

.gitignore
rejoshed: On line 169 `generate-mocks: bin/mockgen generate-deepcopy pkg/mocks/mock_client.go ` generate mocks generates deepcopy as a prerequisite. You shouldn't have to change the EKS-A command. 

Please let me know if you do.

api/v1beta1/cloudstackcluster_webhook_test.go
```diff
@@ -14,198 +14,85 @@ See the License for the specific language governing permissions and
...
+	infrav1 "github.com/aws/cluster-api-provider-cloudstack/api/v1beta1"
+	"github.com/aws/cluster-api-provider-cloudstack/test/dummies"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
 var _ = Describe("CloudStackCluster webhooks", func() {
-	const (
```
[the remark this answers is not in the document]
mrog: Should we be concerned about putting all the test variables in one file?  On one hand, it helps to prevent duplicated names in the same package.  On the other hand, it could turn into a bloated dumping ground for variables.

api/v1beta1/cloudstackcluster_types.go
```diff
@@ -39,13 +39,58 @@ type CloudStackIdentityReference struct {
...
+	for _, zone := range zones {
+		return &zone
+	}
+	return nil
+}
+
+// GetByName fetches a zone by name if present in the map of zone statuses.
+// Needed as there's no short way to do this.
+func (zones ZoneStatusMap) GetByName(name string) *Zone {
+	for zoneName, zone := range zones {
```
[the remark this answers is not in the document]
rejoshed: Because zones is keyed by id.

api/v1beta1/cloudstackcluster_webhook_test.go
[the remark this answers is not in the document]
rejoshed: I think maybe we should be concerned about it becoming a dumping ground, but I think that's about quality -- not whether we should do it or not.

If it does grow much more, we'll want to restructure it. Possibly using sub modules or whatnot. Either way, using test/dummies/vars.go has dropped hundreds of lines from tests that only made testing harder to grok.

We just need to be careful about adding every little thing to the dummies file vs just making small tweaks to existing dummies in the tests themselves.

I'm thinking of them as a launchpad.

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackclusters.yaml
```diff
@@ -72,43 +72,107 @@ spec:
...
+                      type: string
+                    network:
+                      description: The network within the Zone to use.
+                      properties:
+                        id:
+                          description: Cloudstack Network ID the cluster is built
+                            in.
+                          type: string
+                        name:
+                          description: Name of the infrastructure identity to be used.
```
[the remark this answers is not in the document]
rejoshed: These config files are generated from the API dir. I'll see to getting this updated.

hack/nginx.conf
```diff
@@ -43,12 +43,13 @@ http {
 # Add as many IPs as desired.
 stream {
     upstream kubeendpoints {
-        server <server1 IP>:6443 max_fails=3 fail_timeout=10s;
-        server <server2 IP>:6443 max_fails=3 fail_timeout=10s;
+
+        # server <server1 IP>:6443 max_fails=3 fail_timeout=10s;
```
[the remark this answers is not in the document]
rejoshed: These comments were always intended to be human read. That's why <server1 IP> is listed the way it is.

The load balancer won't work without adding upstreams here.

api/v1beta1/cloudstackmachine_types.go
mrog: It's going to take me a while to get used to that.  I've spent many years following Google's Java style guide, which says to use Url and Id.  But the Golang wiki says to use URL and ID.  https://github.com/golang/go/wiki/CodeReviewComments#initialisms

controllers/cloudstackmachine_controller.go
mrog: I disagree that it thwarts the randomness in any meaningful way, at least if the algorithm is implemented correctly.  But it's also completely unnecessary to seed it more than once.  I agree with moving it to main.

mrog: Actually, it can thwart the randomness if you seed it twice in a short enough time that you use the same seed.

controllers/cloudstackmachine_controller.go
```diff
@@ -153,41 +156,55 @@ func (r *CloudStackMachineReconciler) reconcile(
...
+		randNum := (rand.Int() % len(csCluster.Spec.Zones)) // #nosec G404 -- weack crypt rand doesn't matter here.
```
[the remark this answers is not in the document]
jweite-amazon: week -> weak.

api/v1beta1/cloudstackcluster_webhook.go
maxdrib: how do you set the verbosity when running CAPC?

api/v1beta1/cloudstackcluster_types.go
```diff
@@ -65,24 +110,28 @@ type CloudStackClusterSpec struct {
...
-	// The CS zone ID the cluster is built in.
-	ZoneID string `json:"zoneID"`
+	// The status of the cluster's ACS Zones.
+	// +optional
+	Zones ZoneStatusMap `json:"zones,omitempty"`
 
-	// Cloudstack Network ID the cluster is built in.
-	NetworkID string `json:"networkID,omitempty"`
+	// CAPI recognizes failure domains as a method to spread machines.
+	// CAPC sets failure domains per to indicate functionin Zones.
```
maxdrib, line 119: typo: functionin? Not sure what this is supposed to be

hack/add_addresses.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+while true; do
```
maxdrib, line 3: what does this script do? It might be worth adding some comments

hack/nginx.conf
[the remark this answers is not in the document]
maxdrib: They look pretty codey to be human read. Maybe worth adding some additional description surrounding usage, since it's not very clear to newbies like myself

pkg/cloud/tags.go
```diff
@@ -34,71 +39,62 @@ type TagIface interface {
 type ResourceType string
 
 const (
-	clusterTagNamePrefix               = "CAPC_cluster_"
-	createdByCAPCTagName               = "created_by_CAPC"
+	ClusterTagNamePrefix               = "CAPC_cluster_"
+	CreatedByCAPCTagName               = "created_by_CAPC"
```
mrog, line 43: Should these start with upper case letters?  Are you exporting them?

pkg/cloud/network_test.go
```diff
@@ -99,6 +99,15 @@ var _ = Describe("Network", func() {
 		It("correctly identifies an existing network from a network status", func() {
 			Ω(cloud.NetworkExists(dummies.CSCluster.Status.Zones.GetOne().Network)).Should(BeTrue())
 		})
+
+		// It("resolves network details with network ID instead of network name", func() {
```
wongni, line 103: Why is this test case commented?

api/v1beta1/cloudstackmachine_webhook_test.go
```diff
@@ -14,224 +14,90 @@ See the License for the specific language governing permissions and
...
+		})
+
+		It("should reject VM template updates to the CloudStackMachine", func() {
+			dummies.CSMachine1.Spec.Template = "ArbitraryUpdateTemplate"
+			Ω(k8sClient.Update(ctx, dummies.CSMachine1)).
+				Should(MatchError(MatchRegexp(forbiddenRegex, "template")))
+		})
+
+		It("should reject updates to VM details of the CloudStackMachine", func() {
+			dummies.CSMachine1.Spec.Details = map[string]string{"memoryOvercommitRatio": "1.5"}
```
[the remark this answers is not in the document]
rejoshed: Discussed in slack, but basically yes -- it replaces machines with new machine templates. No modification is allowed.

pkg/cloud/tags.go
rejoshed: I use them in tests, but I could also see us using them in the controllers at some point.

hack/add_addresses.sh
rejoshed: Done.

hack/nginx.conf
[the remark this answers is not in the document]
rejoshed: Added additional comment to help clarify.

pkg/cloud/tags.go
mrog: Good reason.  👍

pkg/cloud/network_test.go
rejoshed: This test is covered in other tests now, I just didn't get around to removing it.

Removed. : )

maxdrib (approved):

PR #41, opened 2022-03-09
Change template and offering to explicitly specify ID or Name

wanyufe:
*Issue #, if available:*

*Description of changes:*
Chagne template and offering to explicitly specify ID or Name. 
Before this commit, template and offering is a string which can be a UUID or name, the code will try both when calling CloudStack API. 
After this change, an ID and Name subfields inside template and offering are added. Code will try ID first if ID has a value, otherwise, try name. 
*Testing performed:*


rejoshed (approved):
Please remove zz_generated. Looks good otherwise.

pkg/cloud/instance_test.go
```diff
@@ -132,50 +133,46 @@ var _ = Describe("Instance", func() {
 
 		It("returns errors occuring while fetching sevice offering information", func() {
 			expectVMNotFound()
-			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering).Return("", -1, unknownError)
-			sos.EXPECT().GetServiceOfferingByID(dummies.CSMachine1.Spec.Offering).Return(nil, -1, unknownError)
+			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering.Name).Return("", -1, unknownError)
+			//sos.EXPECT().GetServiceOfferingByID(dummies.CSMachine1.Spec.Offering.ID).Return(nil, -1, unknownError)
```
wongni, line 137: Please remove this line

.gitignore
```diff
@@ -6,7 +6,6 @@
 *.dylib
 bin
 testbin/*
-api/*/zz*generated.deepcopy.go
```
rejoshed, line 9: Why?

wongni (approved):

pkg/cloud/instance_test.go
wanyufe: oh, missed this one.:(

.gitignore
wanyufe: The reason I am removing this from .gitignore is that this generated.deepcopy.go file is needed in eks-a.
eks-a code is referring capc package, and some code (client.object) requires this generated.deepcopy.go code.

CAPC's previous release has this file, after removing this file, eks-a could not build.

Also, other package like vsphere has this generated deepcopy file exposed.

maxdrib (approved):

.gitignore
rejoshed: I don't want to hang up our release, but after the release I would like us to review this again.

api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -63,8 +63,8 @@ func (r *CloudStackMachine) ValidateCreate() error {
 		errorList = append(errorList, field.Forbidden(field.NewPath("spec", "identityRef", "kind"), "must be a Secret"))
 	}
 
-	errorList = webhookutil.EnsureFieldExists(r.Spec.Offering, "Offering", errorList)
-	errorList = webhookutil.EnsureFieldExists(r.Spec.Template, "Template", errorList)
+	errorList = webhookutil.EnsureAtLeastOneFieldExists(r.Spec.Offering.ID, r.Spec.Offering.Name, "Offering", errorList)
```
jweite-amazon, line 66: I imagine that it'd be incorrect to specify both ID and Name.  Should this test be EnsureExactlyOneFieldExists?

wanyufe: In this case, ID takes priority. If both ID and name are both provided, name will be ignored.

We have test cases to cover this. I had a brief discussion with Joshua about similar situation in multiple-zone (both zone id and name provided) and decided that ID will be used in this situation. So offering and template follow the same pattern.

jweite-amazon: I'm not a fan of silently doing something different than what the user requested.  As a user I'd rather be told I'm doing something wrong, rather than pull my hair out trying to figure out why it was behaving differently than I expected.

wanyufe: After thinking it over again, we can add an extra logic when both ID and name provided: 
1. we still use ID, 
2. we check the name provided with name returned from API call by using ID
3. we emit a warning if mismatch, but still continue the flow.

wanyufe (commented):
pushed new version for name mismatch when both ID and name provided for template and offering.

PR #49, opened 2022-03-15
Add multi-zone test with kube-vip

wongni:
*Issue #, if available:*

*Description of changes:*

This PR removes assigning IP addresses to the control plane VMs when a cluster is created in a `shared` network to support more than one control plane nodes in the cluster. The kube-vip handles the resolution of the control plane endpoint IP address using ARP technique. 

*Testing performed:*


test/e2e/common.go
```diff
@@ -286,6 +287,72 @@ func CheckAffinityGroup(clusterName string, affinityType string) []string {
...
+	if err != nil {
+		Fail("Failed to list machines")
+	}
+	cpZoneIdMap := make(map[string]int)
+	mdZoneIdMap := make(map[string]int)
+	zoneIds := []string{}
+
+	for _, vm := range listResp.VirtualMachines {
+		if strings.Contains(vm.Name, clusterName) {
+			By(vm.Name + " is in zone " + vm.Zonename + " (" + vm.Zoneid + ")")
```
mrog, line 304: Thanks for including both the zone name and ID.

test/e2e/common.go
```diff
@@ -286,6 +287,72 @@ func CheckAffinityGroup(clusterName string, affinityType string) []string {
...
+	}
+
+	Expect(len(mdZoneIdMap)).To(Equal(len(zoneNames)))
+	for _, value := range mdZoneIdMap {
+		Expect(value).To(Equal(nWorkerNodesPerZone))
+	}
+	return zoneIds
+}
+
+func checkZoneAssignments(vm *cloudstack.VirtualMachine, cpZoneIdMap map[string]int, mdZoneIdMap map[string]int) error {
```
mrog, line 327: This function can be simplified to this:
```
func checkZoneAssignments(vm *cloudstack.VirtualMachine, cpZoneIdMap map[string]int, mdZoneIdMap map[string]int) {
	if strings.Contains(vm.Name, ControlPlaneIndicator) {
		cpZoneIdMap[vm.Zoneid]++
	}
	
	if strings.Contains(vm.Name, MachineDeploymentIndicator) {
		mdZoneIdMap[vm.Zoneid]++
	}
}
```

test/e2e/multi_zone.go
```diff
@@ -0,0 +1,90 @@
...
+				KubernetesVersion:        input.E2EConfig.GetVariable(KubernetesVersion),
+				ControlPlaneMachineCount: pointer.Int64Ptr(3),
+				WorkerMachineCount:       pointer.Int64Ptr(int64(workerMachineCount)),
+			},
+			WaitForClusterIntervals:      input.E2EConfig.GetIntervals(specName, "wait-cluster"),
+			WaitForControlPlaneIntervals: input.E2EConfig.GetIntervals(specName, "wait-control-plane"),
+			WaitForMachineDeployments:    input.E2EConfig.GetIntervals(specName, "wait-worker-nodes"),
+		}, clusterResources)
+
+		CheckZones(clusterResources.Cluster.Name, []string{"Zone3", "Zone2"}, workerMachineCount)
```
mrog, line 81: These zone names should be configurable using environment variables.

pkg/cloud/instance.go
```diff
@@ -198,14 +197,6 @@ func (c *client) GetOrCreateVMInstance(
...
 	setIfNotEmpty(csCluster.Status.DomainID, p.SetDomainid)
 
-	// If this VM instance is a control plane, consider setting its IP.
-	_, isControlPlanceMachine := capiMachine.ObjectMeta.Labels["cluster.x-k8s.io/control-plane"]
-	if isControlPlanceMachine && zone.Network.Type == NetworkTypeShared {
-		// If the specified control plane endpoint is an IP address, specify the IP address of this VM instance.
-		if net.ParseIP(csCluster.Spec.ControlPlaneEndpoint.Host) != nil {
-			p.SetIpaddress(csCluster.Spec.ControlPlaneEndpoint.Host)
-		}
-	}
```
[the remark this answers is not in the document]
maxdrib: Also would you mind adding more details to the PR description explaining the motivation for this change and what impact it would have? I think this might be a PR we refer back to in the future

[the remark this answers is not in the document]
wongni: Updated the description. 
Even with single CP, I think it's better not to assign IP address to the CP because the kube-vip resolves the control plane endpoint IP address.

maxdrib (approved):

test/e2e/data/infrastructure-cloudstack/v1beta1/bases/cluster-with-kcp.yaml
```diff
@@ -49,8 +49,58 @@ spec:
         name: '{{ local_hostname }}'
         kubeletExtraArgs:
           provider-id: cloudstack:///'{{ ds.meta_data.instance_id }}'
+        ignorePreflightErrors:
+          - DirAvailable--etc-kubernetes-manifests
     preKubeadmCommands:
       - swapoff -a
+    files:
```
wanyufe, line 56: Should we make this kube-vip static-pod conditional? so the customer using external load balancer can choose not using kube-vip.

wongni: This yaml file is only for e2e testing.

rejoshed:
@wongni it's been a bit since this PR has been touched. Are we still moving forward with this?

PR #57, opened 2022-03-30
Support domain ID as well as domain name

wongni:
*Issue #, if available:*

*Description of changes:*
When a sub-domain and its account is used in CloudStackCluster.spec.domain and account capc fails to get a domain ID for the input domain name because GetDomainID doesn't look up the sub-domains. To make capc work with sub-domain listDomains with listall=true should be used. 

*Testing performed:*
Manually created a cluster using eks-a and clusterctl

### DIFF
--- a/api/v1beta1/cloudstackcluster_types.go
+++ b/api/v1beta1/cloudstackcluster_types.go
@@ -101,7 +101,7 @@ type CloudStackClusterSpec struct {
 
 	// CloudStack domain.
 	// +optional
-	Domain string `json:"domain,omitempty"`
+	Domain CloudStackResourceIdentifier `json:"domain,omitempty"`
 
 	// +optional
 	// +k8s:conversion-gen=false
--- a/api/v1beta1/cloudstackcluster_types_test.go
+++ b/api/v1beta1/cloudstackcluster_types_test.go
@@ -49,7 +49,7 @@ var _ = Describe("CloudStackCluster types", func() {
 		})
 
 		It("Should return nil if ZoneStatusMap does not have a zone having the name passed into GetByName", func() {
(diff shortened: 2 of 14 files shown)


rejoshed:
@jweite-amazon 

> Do we need to implement Mikhail's suggestion of supporting domain specification by ID?

image

I think we should do both by ID and or full domain path if possible. So in this case it would be something like root/blah/blah.

I'm not entirely certain that's possible though. Still researching.

wongni:
> Why would we only allow specifying one or the other. It's fine if the user specifies both, they just need to match against the relationship in CloudStack.

I thought name and id are mutually exclusive because eks-a says that. 
```
              template:
                description: Template refers to a VM image template which has been
                  previously registered in CloudStack. It can either be specified
                  as a UUID or name
                properties:
                  id:
                    description: Id of a resource in the CloudStack environment. Mutually
                      exclusive with Name
                    type: string
                  name:
                    description: Name of a resource in the CloudStack environment.
                      Mutually exclusive with Id
                    type: string
```
https://github.com/aws/eks-anywhere/blob/53edfec5399344dc1eadc0b88f661cb068517631/config/crd/bases/anywhere.eks.amazonaws.com_cloudstackmachineconfigs.yaml#L79

api/v1beta1/cloudstackcluster_types.go
```diff
@@ -101,7 +101,7 @@ type CloudStackClusterSpec struct {
 
 	// CloudStack domain.
 	// +optional
-	Domain string `json:"domain,omitempty"`
+	Domain CloudStackResourceIdentifier `json:"domain,omitempty"`
```
maxdrib, line 104: this will be a backwards incompatible change, right? How can we start making these schema changes backwards compatible to decouple ourselves from downstream consumers?

PR #58, opened 2022-03-31
Bug fix: failure to clean-up during CS Machine provisioning if VM provisioned to error state.

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Deleting capiMachine, instead of cloudstackMachine upon detection of VM error state after provisioning.  cloudStackMachine will ultimately be deleted due to its dependency on the capiMachine.

Addressed machine deletion race-condition that occurs when VM is created into error state.

*Testing performed:*
Provision single control plane node cluster specifying a machine offering that the ACS cannot fulfill.  (72 CPUs, 1TB of memory).  This formerly would result in the deletion of the cloudStack machine in question, but not its VM (reconciliation for the cs Machine delete does not get queued), nor its CAPI machine.

Tested the same scenario after fix: capi machine, cs machine and VM all get deleted.  CAPC retries provisioning after the expected delay, and the cycle repeats itself, as reasonable and expected.

### DIFF
--- a/pkg/cloud/instance.go
+++ b/pkg/cloud/instance.go
@@ -216,7 +216,12 @@ func (c *client) GetOrCreateVMInstance(
 		// Regretfully the deployVMResp may be nil, so we need to get the VM ID with a separate query, so we
 		// can return it to the caller, so they can clean it up.
 		listVirtualMachineParams := c.cs.VirtualMachine.NewListVirtualMachinesParams()
+		listVirtualMachineParams.SetTemplateid(templateID)
+		listVirtualMachineParams.SetZoneid(csMachine.Status.ZoneID)
+		listVirtualMachineParams.SetNetworkid(zone.Network.ID)
 		listVirtualMachineParams.SetName(csMachine.Name)
+		setIfNotEmpty(csCluster.Status.DomainID, listVirtualMachineParams.SetDomainid)
+		setIfNotEmpty(csCluster.Spec.Account, listVirtualMachineParams.SetAccount)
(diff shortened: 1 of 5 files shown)


jweite-amazon:
@rejoshed, I discovered a gotcha wrt running controller in a debugger (which is how I was diagnosing this).  When you do so, controller connects to the K8S cluster using your workstation's kubeconfig, and consequently your admin identity.  So the fix I applied worked find when I re-ran in the debugger.  When I subsequently gave it a run in-cluster I discovered the permissions problem.

pkg/cloud/instance.go
```diff
@@ -212,6 +212,14 @@ func (c *client) GetOrCreateVMInstance(
 
 	deployVMResp, err := c.cs.VirtualMachine.DeployVirtualMachine(p)
 	if err != nil {
+		// Just because an error was returned doesn't mean a (failed) VM wasn't created and will need to be dealt with.
+		// Regretfully the deployVMResp may be nil, so we need to get the VM ID with a separate query, so we
+		// can return it to the caller, so they can clean it up.
+		listVirtualMachineParams := c.cs.VirtualMachine.NewListVirtualMachinesParams()
+		listVirtualMachineParams.SetName(csMachine.Name)
```
rejoshed, line 219: Is the name unique enough that we can be sure we're only going to get the one machine back here?

jweite-amazon: I did a preliminary test of attempting to create a VM with the same name as on that existed, and it was rejected.  But in retrospect, it was in the same zone/domain as the existing one, and I can't say that there are circumstance where name along is not sufficient. I'll narrow that query to use as much as we know about the VM we attempted to create.

jweite-amazon:
Narrowed the search for faulty VMs after provisioning failure to include zone, network, domain and account.

wongni (approved):

PR #60, opened 2022-04-01
Fix/add listall list domains

wongni:
*Issue #, if available:*

*Description of changes:*
add listall=true when listing domains to query sub-domains as well

*Testing performed:*
Manually tested

### DIFF
--- a/pkg/cloud/cluster.go
+++ b/pkg/cloud/cluster.go
@@ -106,7 +106,6 @@ func (c *client) ResolveDomainAndAccount(csCluster *infrav1.CloudStackCluster) e
 
 			p.SetListall(true)
 			p.SetName(domainName)
-			p.SetLevel(len(tokens))
 		}
 		resp, retErr := c.cs.Domain.ListDomains(p)
 		if retErr != nil {
--- a/pkg/cloud/cluster.go
+++ b/pkg/cloud/cluster.go
@@ -23,6 +23,8 @@ import (
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+const domainPrefix = "ROOT/"
+
 type ClusterIface interface {
 	GetOrCreateCluster(*infrav1.CloudStackCluster) error
 	DisposeClusterResources(cluster *infrav1.CloudStackCluster) error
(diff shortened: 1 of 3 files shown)


maxdrib (approved):

pkg/cloud/cluster.go
```diff
@@ -69,13 +69,16 @@ func (c *client) GetOrCreateCluster(csCluster *infrav1.CloudStackCluster) (retEr
 
 	// If provided, translate Domain name to Domain ID.
 	if csCluster.Spec.Domain != "" {
```
rejoshed, line 71: Isn't this `Domain.Name` now?

pkg/cloud/cluster.go
```diff
@@ -69,13 +69,16 @@ func (c *client) GetOrCreateCluster(csCluster *infrav1.CloudStackCluster) (retEr
...
-		domainID, count, retErr := c.cs.Domain.GetDomainID(csCluster.Spec.Domain)
+		p := c.cs.Domain.NewListDomainsParams()
+		p.SetListall(true)
+		p.SetName(csCluster.Spec.Domain)
```
rejoshed, line 74: again, `Domain.Name`?

pkg/cloud/cluster.go
```diff
@@ -69,13 +69,16 @@ func (c *client) GetOrCreateCluster(csCluster *infrav1.CloudStackCluster) (retEr
...
 		if retErr != nil {
 			return retErr
-		} else if count != 1 {
-			return errors.Errorf("expected 1 Domain with name %s, but got %d", csCluster.Spec.Domain, count)
+		} else if resp.Count != 1 {
+			return errors.Errorf("expected 1 Domain with name %s, but got %d", csCluster.Spec.Domain, resp.Count)
 		} else {
-			csCluster.Status.DomainID = domainID
+			csCluster.Status.DomainID = resp.Domains[0].Id
 		}
```
rejoshed, line 82: I guess I'm confused as I thought there was already a PR with the Domain.ID and Domain.Name fields added. It also considered logic for checking accounts against domain.

Why this PR?

pkg/cloud/cluster.go
```diff
@@ -92,6 +86,44 @@ func (c *client) GetOrCreateCluster(csCluster *infrav1.CloudStackCluster) (retEr
...
 
+func (c *client) ResolveDomainAndAccount(csCluster *infrav1.CloudStackCluster) error {
+	if (csCluster.Spec.Domain != "" && csCluster.Spec.Account == "") ||
+		(csCluster.Spec.Domain == "" && csCluster.Spec.Account != "") {
+		return errors.Errorf("Both domain and account must be specified or none of them must be specified")
+	}
+
+	if csCluster.Spec.Domain != "" && csCluster.Spec.Account != "" {
+		p := c.cs.Domain.NewListDomainsParams()
+		p.SetListall(true)
```
rejoshed, line 97: Lets at least narrow it by name since it's so easy to do. p.SetName(Domain) or whatever the method is.

pkg/cloud/cluster.go
jweite-amazon: No, that'd break the API, right?  Our goal is to keep it the same as it is today.

pkg/cloud/cluster.go
rejoshed: Otherwise, looks good overall.

wongni: If domain: value is a fqn then setName() make make the list domains return nothing.

rejoshed:
/lgtm

Obviously tests will need to pass, and I suggest something like:
```
tokens := strings.Split("some/path", "/")
p.SetName(tokens[len(tokens)-1])
p.SetLevel(len(tokens))
```

But I won't block anything.

rejoshed (approved):

PR #63, opened 2022-04-06
Use list affinity groups instead of get affinity group by ID or name to support sub-domain

wongni:
*Issue #, if available:*

*Description of changes:*
When create a cluster in a sub-domain with managed affinity group there is an error like this
```
Unable to create affinity group, a group with name ProAffinity-wonkun-management-cluster-7adbe2af-0e59-494b-8c92-cf0f980f2bb6 already exists
```
To fix this, list affinity groups should be used instead of get by ID or name. 

However, this PR does not make this happen because the ACS complains the following error when deploying a VM in a sub-domain with an affinity group created in the sub-domain. 

```
Acct[2e417bec-f13e-4994-9b5e-2d717fd91353-wonkun-admin] does not have permission to operate with resource AffinityGroup
```

*Testing performed:*
Manual testing

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -118,11 +118,6 @@ func (r *CloudStackMachineReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return reconcile.Result{}, nil
 	}
 
-	// Delete VM instance if deletion timestamp present.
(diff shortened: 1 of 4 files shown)


rejoshed:
@wongni can we close this one too? I'm doing all these things in my subdomain user operations PR.

PR #67, opened 2022-04-21
Add diskOffering to machine template config

wanyufe:
*Issue #, if available:*
Feature add: allow customer to optionally specify disk offering ID/Name in machine template config. 
*Description of changes:*
Add diskOffering to machine template config
*Testing performed:*
Unit testing.


api/v1beta1/cloudstackmachine_types.go
```diff
@@ -43,6 +43,9 @@ type CloudStackMachineSpec struct {
 	// CloudStack template to use.
 	Template CloudStackResourceIdentifier `json:"template"`
 
+	// CloudStack disk offering to use.
+	DiskOffering CloudStackResourceDiskOffering `json:"diskOffering,omitempty"`
```
maxdrib, line 47: should this be a list, in case a customer wants to add multiple disk offerings?

wanyufe: According to https://cloudstack.apache.org/api/apidocs-4.14/apis/deployVirtualMachine.html
only one diskofferingId is supported when deploy a VM in cloudstack.

That being said, above api has an alternative parameter `datadiskofferinglist` which can provider an list of disk template to disk offering id map. But `datadiskofferinglist` is designed to a special template which supports multiple disk OVA file)

I did not find any info that qcow2 in cloudstack support multiple data disk, - can confirm with cloudstack experts later.

In the future, if multiple data disk template is supported, the proper places to configure `datadiskofferinglist` should be under templateOffering section since it is template specific and data disk could not be customized.

api/v1beta1/cloudstackmachinetemplate_webhook.go
```diff
@@ -97,6 +97,7 @@ func (r *CloudStackMachineTemplate) ValidateUpdate(old runtime.Object) error {
 
 	errorList := field.ErrorList(nil)
 	errorList = webhookutil.EnsureBothFieldsAreEqual(spec.Offering.ID, spec.Offering.Name, oldSpec.Offering.ID, oldSpec.Offering.Name, "offering", errorList)
+	errorList = webhookutil.EnsureBothFieldsAreEqual(spec.DiskOffering.ID, spec.DiskOffering.Name, oldSpec.DiskOffering.ID, oldSpec.DiskOffering.Name, "diskOffering", errorList)
 	errorList = webhookutil.EnsureStringFieldsAreEqual(spec.SSHKey, oldSpec.SSHKey, "sshkey", errorList)
```
wongni, line 101: Is checking DiskOffering.MountPath missing here?

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackmachines.yaml
```diff
@@ -76,6 +76,20 @@ spec:
                   type: string
                 description: Optional details map for deployVirtualMachine
                 type: object
+              diskOffering:
+                description: CloudStack disk offering to use.
+                properties:
+                  id:
+                    description: Cloudstack resource ID.
+                    type: string
+                  mountPath:
```
wongni, line 85: description is missing

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackmachinetemplates.yaml
```diff
@@ -62,6 +62,20 @@ spec:
                           type: string
                         description: Optional details map for deployVirtualMachine
                         type: object
+                      diskOffering:
+                        description: CloudStack disk offering to use.
+                        properties:
+                          id:
+                            description: Cloudstack resource ID.
+                            type: string
+                          mountPath:
```
wongni, line 71: description is missing

api/v1beta1/cloudstackmachinetemplate_webhook.go
wanyufe: mountPath is not used here since disk is not mount by capc, but by eks-a script.

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackmachinetemplates.yaml
wanyufe: adding description.

api/v1beta1/cloudstackmachine_types.go
```diff
@@ -85,6 +88,18 @@ type CloudStackResourceIdentifier struct {
...
+type CloudStackResourceDiskOffering struct {
+	CloudStackResourceIdentifier `json:",inline"`
+	// mount point the data disk uses to mount. The actual partition, mkfs and mount are done by cloud-init generated by kubeadmConfig.
+	MountPath string `json:"mountPath"`
+	// device name of data disk
+	Device string `json:"device"`
+	// filesystem used by data disk
+	Filesystem string `json:"filesystem"`
+	// label of data disk
+	Label string `json:"label"`
```
maxdrib, line 100: are all of these attributes required? it seems like device name and label might be mutually exclusive, no? Maybe some more detail would be helpful in the descriptions, since I'm not really sure why a disk offering would need a "label" or "device name".

test/dummies/vars.go
```diff
@@ -160,6 +169,15 @@ func SetDummyCSMachineVars() {
...
 				Name: "Offering",
 			},
+			DiskOffering: capcv1.CloudStackResourceDiskOffering{
+				CloudStackResourceIdentifier: capcv1.CloudStackResourceIdentifier{
+					Name: "DiskOffering",
+				},
+				MountPath:  "/data",
+				Device:     "/dev/vdb",
+				Filesystem: "ext4",
+				Label:      "data_disk",
```
maxdrib, line 179: Should we have more validation on these fields to make sure they're valid? At least to make sure that "filesystem" is properly input as a courtesy to users who may have made a typo?

api/v1beta1/cloudstackmachine_types.go
```diff
@@ -85,6 +88,18 @@ type CloudStackResourceIdentifier struct {
...
 }
 
+type CloudStackResourceDiskOffering struct {
+	CloudStackResourceIdentifier `json:",inline"`
+	// mount point the data disk uses to mount. The actual partition, mkfs and mount are done by cloud-init generated by kubeadmConfig.
+	MountPath string `json:"mountPath"`
+	// device name of data disk
+	Device string `json:"device"`
+	// filesystem used by data disk
+	Filesystem string `json:"filesystem"`
```
maxdrib, line 98: Where are all these fields used? I don't see them being passed to the CloudStack API anywhere

maxdrib:
Looks like some prow tests are failing

api/v1beta1/cloudstackmachine_types.go
wanyufe: I updated the comments for those new fields. Basically, 
device is the device name used when parition a disk, like /dev/vdb which is a device name in linux.
filesystem is the file system used when mkfs (format) a partition, like ext4, xfs
label is a parameter when format disk (mkfs) and label a partition, it is a recommended way to use when mount fs into a directory in linux.

That being said, other than the disk offering ID/Name which is used by CAPC when deploy VM, other parameters are not used by CAPC (device, filesystem, mountPath, label) those parameters are used by kubeadmconfig spec to generate cloud-init script. 

The reason why these parameters are added in cloudstack machine is that eks-a needs to use cloudstack machine spec to re-construct eks-a template to determine if a upgrade is needed. Without these parameters added in CAPC, eks-a will not know if disk offering config changes or not, and could not make a decision to upgrade. 

That being said, if customer uses CAPC directly without eks-a, the deployed VM will have a data disk available when booting, but no partition or file system applied to data disk, and nothing is mounted.

Let CAPC to parition disk, install filesystem and mount to a directory is an alternative, and our implementation uses kubeadmconfig. 

Here is a proposal in 2020 about similar requirements  
https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200423-etcd-data-disk.md

test/dummies/vars.go
wanyufe: those fields are not used by CAPC directly and verification logic is inside eks-a. See above comments for background.

api/v1beta1/cloudstackmachine_types.go
wanyufe: See above comments.

api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -82,6 +82,11 @@ func (r *CloudStackMachine) ValidateUpdate(old runtime.Object) error {
 	oldSpec := oldMachine.Spec
 
 	errorList = webhookutil.EnsureBothFieldsAreEqual(r.Spec.Offering.ID, r.Spec.Offering.Name, oldSpec.Offering.ID, oldSpec.Offering.Name, "offering", errorList)
+	errorList = webhookutil.EnsureBothFieldsAreEqual(r.Spec.DiskOffering.ID, r.Spec.DiskOffering.Name, oldSpec.DiskOffering.ID, oldSpec.DiskOffering.Name, "diskOffering", errorList)
```
rejoshed, line 85: I'll remove this comment if I find the reason in my readthrough, but why would we require name and ID to be the same?

Does a disk offering not have a name and ID?

pkg/cloud/gomock_reflect_1599216259/prog.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
```
rejoshed, line 5: Pretty sure we don't need this file here. It's generated for Mocks right?

If I'm correct, please remove it and add a .gitignore entry. Something like `pkg/cloud/gomock_reflect_*`

pkg/cloud/helpers.go
```diff
@@ -23,13 +23,20 @@ import (
...
+type setArray func([]string)
 
 func setIfNotEmpty(str string, setFn set) {
 	if str != "" {
 		setFn(str)
 	}
 }
 
+func setArrayIfNotEmpty(strArray []string, setFn setArray) {
+	if len(strArray) > 0 {
```
rejoshed, line 35: These little funcs are fun aren't they!

pkg/cloud/instance.go
```diff
@@ -139,6 +138,37 @@ func (c *client) ResolveTemplate(
...
+			return "", multierror.Append(retErr, errors.Wrapf(
+				err, "could not get DiskOffering by ID %s", csMachine.Spec.DiskOffering.ID))
+		} else if count != 1 {
+			return "", multierror.Append(retErr, errors.Errorf(
+				"expected 1 DiskOffering with UUID %s, but got %d", csMachine.Spec.DiskOffering.ID, count))
+		}
+
+		if len(csMachine.Spec.DiskOffering.Name) > 0 && csMachine.Spec.DiskOffering.Name != csDiskOffering.Name {
+			return "", multierror.Append(retErr, errors.Errorf(
+				"diskOffering name %s does not match name %s returned using UUID %s", csMachine.Spec.DiskOffering.Name, csDiskOffering.Name, csMachine.Spec.DiskOffering.ID))
```
rejoshed, line 154: We don't have a line length enforcing linter in place yet, but could you please add a or many newlines to this one to make it a bit easier to read?

pkg/cloud/instance.go
```diff
@@ -161,13 +191,18 @@ func (c *client) GetOrCreateVMInstance(
 	if err != nil {
 		return err
 	}
+	diskOfferingID, err := c.ResolveDiskOffering(csMachine)
```
rejoshed, line 194: Here we're setting the disk offering on the csMachine and returning it. Maybe we can drop the return and just set it? Most of the methods with Resolve in the name operate on the passed variable only.

pkg/cloud/instance.go
```diff
@@ -139,6 +138,37 @@ func (c *client) ResolveTemplate(
 	return templateID, nil
 }
 
+func (c *client) ResolveDiskOffering(csMachine *infrav1.CloudStackMachine) (diskOfferingID string, retErr error) {
```
rejoshed, line 141: Please add a comment above all functions/methods that are exported.

``` Go
\\ ResolveDiskOffering resolves to make disk offerrings the best thing ever!
func (c *client) ResolveDiskOffering(csMachine *infrav1.CloudStackMachine) (diskOfferingID string, retErr error) {
...
```
There's a linter for this somewhere, but I guess we've not got it turned  on as we still have quite a few without comments.

Personally, I put them there whether they're exported or not. : )

rejoshed (changes requested):
Listed a few requests for change.

Otherwise looks pretty good!

pkg/cloud/instance.go
```diff
@@ -231,13 +266,49 @@ func (c *client) DestroyVMInstance(csMachine *infrav1.CloudStackMachine) error {
...
+	setArrayIfNotEmpty(volIDs, p.SetVolumeids)
 	if _, err := c.csAsync.VirtualMachine.DestroyVirtualMachine(p); err != nil &&
 		strings.Contains(strings.ToLower(err.Error()), "unable to find uuid for id") {
 		// VM doesn't exist. Success...
 		return nil
 	} else if err != nil {
 		return err
 	}
+
+	if err := c.ResolveVMInstanceDetails(csMachine); err == nil && (csMachine.Status.InstanceState == "Expunging" ||
```
rejoshed, line 283: I'm pretty sure these changes are already in main? I think if you click edit and then change base, then don't really change the base it will refresh the diff?

rejoshed: Cool, either way this branch is going to need some merging with main after the CRD branch goes through. Why don't we take another look then.

rejoshed (commented):
I approve, but I'm not going to lgtm it yet. I want to have this merge after the CRD breakout merge.

rejoshed (approved):

rejoshed:
Hey Yufei... 

You can add the lgtm label yourself via the labels button in the PR.

But, you cannot add it via / lgtm.

Anyway, it wouldn't matter as you do need to merge to master first. Let me know if I can help out with that.

PR #70, opened 2022-04-27
Adding support for deletion when a VM is detected as expunging

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR modifies the VM deletion behavior so that if a CloudStack VM is expunging or expunged, it will be considered "deleted" from the CAPC perspective

*Testing performed:*
Ginkgo unit tests pass
TODO: manual e2e testing with clusterctl/tilt

### DIFF
--- a/pkg/cloud/client.go
+++ b/pkg/cloud/client.go
@@ -58,8 +58,8 @@ func NewClient(ccPath string) (Client, error) {
 		return nil, errors.Wrapf(err, "error encountered while parsing [Global] section from config at path: %s", ccPath)
 	}
 
-	c.cs = cloudstack.NewAsyncClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
-	c.csAsync = cloudstack.NewClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
+	c.csAsync = cloudstack.NewAsyncClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
+	c.cs = cloudstack.NewClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
 	_, err := c.cs.APIDiscovery.ListApis(c.cs.APIDiscovery.NewListApisParams())
 	if err != nil && strings.Contains(strings.ToLower(err.Error()), "i/o timeout") {
 		return c, errors.Wrap(err, "timeout while checking CloudStack API Client connectivity")
@@ -68,6 +68,6 @@ func NewClient(ccPath string) (Client, error) {
 }
 
 func NewClientFromCSAPIClient(cs *cloudstack.CloudStackClient) Client {
-	c := &client{cs: cs}
+	c := &client{cs: cs, csAsync: cs}
 	return c
 }
--- a/pkg/cloud/client.go
+++ b/pkg/cloud/client.go
@@ -58,8 +58,12 @@ func NewClient(ccPath string) (Client, error) {
 		return nil, errors.Wrapf(err, "error encountered while parsing [Global] section from config at path: %s", ccPath)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cloud/client.go
```diff
@@ -58,8 +58,8 @@ func NewClient(ccPath string) (Client, error) {
 		return nil, errors.Wrapf(err, "error encountered while parsing [Global] section from config at path: %s", ccPath)
 	}
 
-	c.cs = cloudstack.NewAsyncClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
-	c.csAsync = cloudstack.NewClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
+	c.csAsync = cloudstack.NewAsyncClient(cfg.APIURL, cfg.APIKey, cfg.SecretKey, cfg.VerifySSL)
```
wongni, line 61: The client returned from NewAsyncClient works in a synchronous way. On the other hand, a client returned from NewClient works in an asychronous way. I know it's confusing but that's the way it is. Please revert back this change or we have to change all places using csAsync and cs.

pkg/cloud/instance.go
```diff
@@ -239,5 +238,18 @@ func (c *client) DestroyVMInstance(csMachine *infrav1.CloudStackMachine) error {
 	} else if err != nil {
 		return err
 	}
+
+	if err := c.ResolveVMInstanceDetails(csMachine); err == nil && csMachine.Status.InstanceState == "Expunging" ||
```
wongni, line 242: Don't we need () around `csMachine.Status.InstanceState == "Expunging" || csMachine.Status.InstanceState == "Expunged"`

pkg/cloud/instance_test.go
```diff
@@ -325,4 +325,57 @@ var _ = Describe("Instance", func() {
...
+			vms.EXPECT().NewDestroyVirtualMachineParams(*dummies.CSMachine1.Spec.InstanceID).
+				Return(expungeDestroyParams)
+			vms.EXPECT().DestroyVirtualMachine(expungeDestroyParams).Return(nil, nil)
+			vms.EXPECT().GetVirtualMachinesMetricByID(*dummies.CSMachine1.Spec.InstanceID).Return(nil, -1, notFoundError)
+			vms.EXPECT().GetVirtualMachinesMetricByName(dummies.CSMachine1.Name).Return(nil, -1, notFoundError)
+			Ω(client.DestroyVMInstance(dummies.CSMachine1)).
+				Should(Succeed())
+		})
+
+		It("calls destroy without error and identifies it as expunging", func() {
```
wongni, line 358: Please add a test case with Expunged state

pkg/cloud/client.go
maxdrib: Got it - I wasn't sure about this so I'm glad you caught it. I'll leave a comment to that effect to avoid future confusion. Could you share a link to some documentation.

pkg/cloud/instance.go
maxdrib: Yes, good catch!

pkg/cloud/instance_test.go
maxdrib: Will do

pkg/cloud/client.go
rejoshed: https://pkg.go.dev/github.com/apache/cloudstack-go/v2/cloudstack#NewAsyncClient

It's annoying to me too.

> For sync API calls this client behaves exactly the same as a standard client call, but for async API calls this client will wait until the async job is finished or until the configured AsyncTimeout is reached. When the async job finishes successfully it will return actual object received from the API and nil, but when the timout is reached it will return the initial object containing the async job ID for the running job and a warning.

rejoshed:
lgtm to me, but not adding the label. Will let Wonkun or Max do that. : )

rejoshed:
Max, wasn't sure if you know, but if you want to add the label yourself do so as shown in the image.

image

rejoshed (approved):

wongni (approved):

PR #73, opened 2022-05-02
Add moving the cluster to the workload cluster to make the cluster a management cluster

wongni:
*Issue #, if available:*

*Description of changes:*

Adding an e2e test case that turning a workload cluster to a management cluster. 
Note that this test doesn't work with the current capc controller image building and pushing because we uses a local repository for storing the image, however, the image in the local repository is not accessible from the workload cluster. To run this test we need to push to a remote repository so that the image is reachable from both the bootstrap and the workload clusters. 

```
IMG=public.ecr.aws/r9t0b9f2/cluster-api-provider-capc:e2e-20220502a make docker-build docker-push
JOB=self-hosted make IMG_LOCAL=public.ecr.aws/r9t0b9f2/cluster-api-provider-capc:e2e-20220502a run-e2e
```

*Testing performed:*

### DIFF
--- a/test/e2e/self_hosted.go
+++ b/test/e2e/self_hosted.go
@@ -1,0 +1,206 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


maxdrib:
This would be a really great test case to include in our suite. Is there any way we can override the image uri to use some public ECR uri instead of localhost?

PR #77, opened 2022-05-09
Bumping ginkgo to v2 and adding code coverage check

maxdrib:
*Issue #, if available:*

*Description of changes:*
This PR bumps ginkgo to version 2 and gets rid of the annoying ginkgo deprecation notice. Also, it adds a github actions workflow to upload codecoverage reports so we can observe our test coverage. Have a look: https://codecov.io/gh/aws/cluster-api-provider-cloudstack/tree/0522763178ecbde09a4b09fb34926a10a058b74c

Inspired by EKS-A's codecov workflow: https://github.com/aws/eks-anywhere/pull/1914

*Testing performed:*
Unit tests pass locally


rejoshed:
/approve

Looks like you're having troubles with the junit reporter though.

Also, why don't we use Prow for this instead of GH Actions? I mean, I _love_ GH Actions, but it seems odd that we'd split the two CI systems.

I can help you convert to Prow if you'd like?

api/v1beta1/webhook_suite_test.go
```diff
@@ -49,9 +48,8 @@ var cancel context.CancelFunc
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Webhook Suite",
-		[]Reporter{printer.NewlineReporter{}})
+	RunSpecs(t,
```
wongni, line 51: Please make this a single line

controllers/controllers_suite_test.go
```diff
@@ -100,9 +100,8 @@ func TestAPIs(t *testing.T) {
 
 	RegisterFailHandler(Fail)
 
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Controller Suite",
-		[]Reporter{printer.NewlineReporter{}})
+	RunSpecs(t,
```
wongni, line 103: Same here

wongni (changes requested):

wongni (approved):

api/v1beta1/webhook_suite_test.go
```diff
@@ -66,7 +64,15 @@ var _ = BeforeSuite(func() {
 		},
 	}
 
-	cfg, err := testEnv.Start()
+	var cfg *rest.Config
```
rejoshed, line 67: Hmm... V2 requires an explicit GinkgoRecover here where it wasn't required prior?

maxdrib: Correct, see the discussion in https://github.com/onsi/ginkgo/issues/882. Basically it sounds like this is how we can enforce a timeout on an async test

PR #85, opened 2022-05-12
Tests/Adhoc Domain/Account Creation and Test Isolation

rejoshed:
*Issue #, if available:*
N/A

*Description of changes:*
Adds Domain/Account/User creation and cleanup on unit test runs.

The unit tests now still only take about 15 seconds, but now they run in a domain with UUID that is cleaned up afterwards. This allows test isolation and helps to guarantee failed test cleanup.

In doing so I've also managed to turn enable about 40+ unit tests that were skipped in the pipeline previously.

This PR also moves unit tests to their own module/folder. This allows tools like dep check to consider our go.mod and go.sum without additional back bending to remove test only requirements.

*Testing performed:*
More than before!


test/unit/cloud/user_credentials_test.go
```diff
@@ -37,21 +37,16 @@ var _ = Describe("User Credentials", func() {
...
-			if connectionErr != nil { // Only do these tests if an actual ACS instance is available via cloud-config.
-				Skip(errors.Wrapf(connectionErr, "Could not connect to ACS instance").Error())
-			}
-
-			// Settup dummies.
-			// TODO: move these to the test dummies package.
-			domain = cloud.Domain{Path: "ROOT/blah/blah/subsub"}
-			account = cloud.Account{Name: "SuperNested", Domain: domain}
-			user = cloud.User{Name: "SubSub", Account: account}
+			client = realCloudClient
```
rejoshed, line 45: All of this cloud client setup stuff has now moved to the BeforeSuite setup.

PR #89, opened 2022-05-13
add support for customized disk offering

wanyufe:
*Issue #, if available:*
Add support for customized diskoffering, a new attribue `sizeInGB` is added under diskoffering.
*Description of changes:*
CAPC (instance.go) is going to check if provided diskoffering is customized offering not, only customized disk offering supports sizeInGB with int64 > 0 value.
Error will happen if:

1. customized disk offering with sizeInGB == 0
2. non-customized disk offering with sizeInGB > 0

*Testing performed:*
Unit testing.


pkg/cloud/helpers.go
```diff
@@ -37,6 +38,12 @@ func setArrayIfNotEmpty(strArray []string, setFn setArray) {
 	}
 }
 
+func setIntIfNotZero(num int64, setFn setInt) {
```
mrog, line 41: Maybe rename this to setIntIfGreaterThanZero.

wongni: How about `setIfPositiveNumber`

pkg/cloud/instance.go
```diff
@@ -139,35 +139,48 @@ func (c *client) ResolveTemplate(
...
 		return "", multierror.Append(retErr, errors.Wrapf(
-			err, "could not get DiskOffering ID from %s", csMachine.Spec.DiskOffering.Name))
+			err, "could not get DiskOffering by ID %s", diskOfferingID))
 	} else if count != 1 {
 		return "", multierror.Append(retErr, errors.Errorf(
-			"expected 1 DiskOffering with name %s, but got %d", csMachine.Spec.DiskOffering.Name, count))
+			"expected 1 DiskOffering with UUID %s, but got %d", diskOfferingID, count))
+	}
+
+	if csDiskOffering.Iscustomized && csMachine.Spec.DiskOffering.Size == 0 {
```
wongni, line 172: Can `csMachine.Spec.DiskOffering.Size` be zero at this point? Looks like the webhooks make sure the size be positive, right?

wongni (approved):

api/v1beta1/cloudstackmachine_webhook_test.go
```diff
@@ -46,6 +46,16 @@ var _ = Describe("CloudStackMachine webhook", func() {
 			Expect(k8sClient.Create(ctx, dummies.CSMachine1)).Should(Succeed())
 		})
 
+		It("should accept a CloudStackMachine with disk Offering size attribute", func() {
+			dummies.CSMachine1.Spec.DiskOffering = dummies.DiskOffering3
```
rejoshed, line 50: Might want to rename DiskOffering3 to CustomSizeDiskOffering to differentiate.

api/v1beta1/cloudstackmachine_webhook_test.go
```diff
@@ -46,6 +46,16 @@ var _ = Describe("CloudStackMachine webhook", func() {
...
+			Expect(k8sClient.Create(ctx, dummies.CSMachine1)).Should(Succeed())
+		})
+
+		It("should not accept a CloudStackMachine with disk Offering size attribute", func() {
```
rejoshed, line 54: The `it` here and the `it` test above this appear to contradict each other?

```"should accept a CloudStackMachine with disk Offering size attribute"```

```"should not accept a CloudStackMachine with disk Offering size attribute"```

File: api/v1beta1/cloudstackmachine_webhook_test.go [hunk 5]
```diff
@@ -46,6 +46,16 @@ var _ = Describe("CloudStackMachine webhook", func() {
...
+			dummies.CSMachine1.Spec.DiskOffering = dummies.DiskOffering4
```
rejoshed, line 55: I'm guessing this test is about putting a bad attribute into the test, but DiskOffering4 isn't very descriptive either.

Another thing we could do, instead of having so many disk offering dummy vars is just set the single relevant attribute for the `It` block.

The dummy var attributes are reset for every `it` block.

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackmachines.yaml
```diff
@@ -116,6 +116,11 @@ spec:
               diskOffering:
                 description: CloudStack disk offering to use.
                 properties:
+                  customSizeInGB:
+                    description: disk size if disk offering is custom disk (need to
```
rejoshed, line 120: I think here

> disk size if disk offering is custom disk (need to...

You actually mean

> disk size if disk offering is custom disk (needed to...

adding the `ed` on on `need`

config/crd/bases/infrastructure.cluster.x-k8s.io_cloudstackmachinetemplates.yaml
```diff
@@ -103,6 +103,11 @@ spec:
                       diskOffering:
                         description: CloudStack disk offering to use.
                         properties:
+                          customSizeInGB:
+                            description: disk size if disk offering is custom disk
```
rejoshed, line 107: I think we should align on whether we do full sentences with punctuation in descriptions or not.

Just above this description we did use a proper sentence and here we did not. I think we should use full sentences, but I'm mostly concerned with consistency.

controllers/controllers_suite_test.go
```diff
@@ -172,7 +171,7 @@ var _ = BeforeSuite(func() {
 		K8sClient:  k8sManager.GetClient(),
 		Scheme:     k8sManager.GetScheme(),
 		CSClient:   CS,
-		BaseLogger: logr.New(logf.NullLogSink{})}
+		BaseLogger: logr.Discard()}
```
rejoshed, line 174: Interesting. Could you please give me some background on why you changed this?

pkg/cloud/helpers.go
```diff
@@ -37,6 +38,12 @@ func setArrayIfNotEmpty(strArray []string, setFn setArray) {
 	}
 }
 
+func setIntIfPositive(num int64, setFn setInt) {
```
rejoshed, line 41: Look at us go!

I bet generics will make this sort of things much easier someday.

pkg/cloud/instance.go
```diff
@@ -139,35 +139,48 @@ func (c *client) ResolveTemplate(
 // disk offering name matches name provided in spec.
 // If disk offering ID is not provided, the disk offering name is used to retrieve disk offering ID.
 func (c *client) ResolveDiskOffering(csMachine *infrav1.CloudStackMachine) (diskOfferingID string, retErr error) {
-	if len(csMachine.Spec.DiskOffering.ID) > 0 {
-		csDiskOffering, count, err := c.cs.DiskOffering.GetDiskOfferingByID(csMachine.Spec.DiskOffering.ID)
+	diskOfferingID = csMachine.Spec.DiskOffering.ID
```
rejoshed, line 142: I'm wondering if maybe we should break DiskOfferings into their own file and Interface?

I'm fine either way, but I know there's a lot going on in the Instance interface and file already.

pkg/cloud/instance.go
```diff
@@ -139,35 +139,48 @@ func (c *client) ResolveTemplate(
...
+		diskID, count, err := c.cs.DiskOffering.GetDiskOfferingID(csMachine.Spec.DiskOffering.Name)
 		if err != nil {
 			return "", multierror.Append(retErr, errors.Wrapf(
-				err, "could not get DiskOffering by ID %s", csMachine.Spec.DiskOffering.ID))
+				err, "could not get DiskOffering ID from %s", csMachine.Spec.DiskOffering.Name))
 		} else if count != 1 {
 			return "", multierror.Append(retErr, errors.Errorf(
-				"expected 1 DiskOffering with UUID %s, but got %d", csMachine.Spec.DiskOffering.ID, count))
+				"expected 1 DiskOffering with name %s, but got %d", csMachine.Spec.DiskOffering.Name, count))
+		} else {
```
rejoshed, line 151: Shouldn't this just be another `else if`?

pkg/cloud/instance.go
```diff
@@ -139,35 +139,48 @@ func (c *client) ResolveTemplate(
...
-		if len(csMachine.Spec.DiskOffering.Name) > 0 && csMachine.Spec.DiskOffering.Name != csDiskOffering.Name {
-			return "", multierror.Append(retErr, errors.Errorf(
-				"diskOffering name %s does not match name %s returned using UUID %s",
-				csMachine.Spec.DiskOffering.Name, csDiskOffering.Name, csMachine.Spec.DiskOffering.ID))
-		}
-		return csMachine.Spec.DiskOffering.ID, nil
+		diskOfferingID = diskID
 	}
-	if len(csMachine.Spec.DiskOffering.Name) == 0 {
+	if len(diskOfferingID) == 0 {
```
rejoshed, line 160: Should this return nil here? It feels like if the disk offering name was provided, but the ID can't be found, we should return an error or something.

pkg/cloud/instance.go
```diff
@@ -139,35 +139,48 @@ func (c *client) ResolveTemplate(
...
 		return "", multierror.Append(retErr, errors.Wrapf(
-			err, "could not get DiskOffering ID from %s", csMachine.Spec.DiskOffering.Name))
+			err, "could not get DiskOffering by ID %s", diskOfferingID))
 	} else if count != 1 {
 		return "", multierror.Append(retErr, errors.Errorf(
-			"expected 1 DiskOffering with name %s, but got %d", csMachine.Spec.DiskOffering.Name, count))
+			"expected 1 DiskOffering with UUID %s, but got %d", diskOfferingID, count))
+	}
+
+	if csDiskOffering.Iscustomized && csMachine.Spec.DiskOffering.CustomSize == 0 {
```
rejoshed, line 172: TBH, I think this method just needs to be broken into smaller methods. There's just too much happening in this one method.

pkg/cloud/instance_test.go
```diff
@@ -179,6 +179,56 @@ var _ = Describe("Instance", func() {
 				ShouldNot(Succeed())
 		})
 
+		It("returns errors when more than one diskoffering found", func() {
+			expectVMNotFound()
+			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering.Name).
+				Return(dummies.CSMachine1.Spec.Offering.ID, 1, nil)
+			ts.EXPECT().GetTemplateID(dummies.CSMachine1.Spec.Template.Name, executableFilter, dummies.Zone1.ID).Return(dummies.CSMachine1.Spec.Template.ID, 1, nil)
+			dos.EXPECT().GetDiskOfferingID(dummies.CSMachine1.Spec.DiskOffering.Name).Return(diskOfferingFakeID, 2, nil)
+			//dos.EXPECT().GetDiskOfferingByID(dummies.CSMachine1.Spec.DiskOffering.ID).Return(&cloudstack.DiskOffering{Iscustomized: false}, 1, nil)
```
rejoshed, line 188: Please remove commented out mock expectation.

rejoshed (changes requested):

rejoshed (approved):

PR #92, opened 2022-05-16
Fix CloudStackMachine Missing Finalizer

rejoshed:
*Issue #, if available:* N/A

*Description of changes:*
If creating a VM returns an error, then the CSMachine finalizer isn't added to the CR.

When the CAPI machine is deleted, the CSMachine gets garbage collected without deleting a machine.

It's a bit of a strange paradigm that a VM gets created yet the creation method returns an error, but that's the case.

May want to by default just add finalizers in controllers in the future.

*Testing performed:*
Used an `Insane Instance` to incite an error on VM creation.

Watched as CAPC cycled VMs several times without continually creating them.
image
image
image

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -193,8 +193,9 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes 
 
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
 		r.Log.Info("CloudStack instance Created", "instanceStatus", r.ReconciliationSubject.Status)
-		controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
 	}
+	// Always add the finalizer regardless. It can't be added twice anyway.
+	controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
 	return ctrl.Result{}, err
 }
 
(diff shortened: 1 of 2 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -193,8 +193,9 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes
 
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
```
rejoshed, line 194: If there was ever an error in VM creation, the finalizer was then not added.

controllers/cloudstackmachine_controller.go
```diff
@@ -204,7 +205,7 @@ func (r *CloudStackMachineReconciliationRunner) RequeueIfInstanceNotRunning() (r
 		r.Log.Info("Machine instance is Running...")
 		r.ReconciliationSubject.Status.Ready = true
 	} else if r.ReconciliationSubject.Status.InstanceState == "Error" {
-		r.Log.Info("CloudStackMachine VM in error state. Deleting associated Machine.", "csMachine", r.ReconciliationSubject)
+		r.Log.Info("CloudStackMachine VM in error state. Deleting associated Machine.", "csMachine", r.ReconciliationSubject.GetName())
```
rejoshed, line 208: We were logging just too darn much info here. Name is a bit less verbose.

controllers/cloudstackmachine_controller.go
```diff
@@ -259,6 +260,7 @@ func (r *CloudStackMachineReconciliationRunner) ReconcileDelete() (retRes ctrl.R
 		}
 		return ctrl.Result{}, err
 	}
+	r.Log.Info("VM Deleted.")
```
rejoshed, line 263: Helpful to know when a machine was in fact deleted.

controllers/utils/utils.go
```diff
@@ -143,20 +143,20 @@ func GetOwnerOfKind(ctx context.Context, c clientPkg.Client, owned clientPkg.Obj
 		return errors.Errorf(
 			"found more than one GVK for owner when finding owner kind for %s/%s", owned.GetName(), owned.GetNamespace())
 	}
-	kind := gvks[0].Kind
+	gvk := gvks[0]
+
 	for _, ref := range owned.GetOwnerReferences() {
-		if ref.Kind != kind {
+		if ref.Kind != gvk.Kind {
```
rejoshed, line 149: These are just log message fixups that I noticed while troubleshooting.

maxdrib (approved):
Great catch! How do we start working on unit tests to make sure these edge cases are covered?

PR #121, opened 2022-06-16
Tests/Account & Domain Isolation and More Domain/Account Tests

rejoshed:
Issue #, if available:
N/A

Description of changes:
Adds Domain/Account/User creation and cleanup on unit test runs.

The unit tests now still only take about 15 seconds, but now they run in a domain with UUID that is cleaned up afterwards. This allows test isolation and helps to guarantee failed test cleanup.

In doing so I've also managed to turn enable about 40+ unit tests that were skipped in the pipeline previously.

Sigh, and now they're still skipped since we don't yet have an ACS instance to hit.

I labeled the integ tests and filtered them in the makefile for now.

Testing performed:
More than before!
image


Makefile
```diff
@@ -23,7 +23,7 @@ RELEASE_DIR ?= out
 export ACK_GINKGO_DEPRECATIONS := 1.16.5
 export ACK_GINKGO_RC=true
 
-PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+export PROJECT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
rejoshed, line 26: Realpath is actually present in our test cointainers.

.golangci.yml
```diff
@@ -33,3 +33,12 @@ run:
...
     - test
+
+issues:
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - dupl
+        - gosec
```
rejoshed, line 44: Go sec was complaining about exec-ing a local script in a test. Just removed it from linting our test code.

Makefile
```diff
@@ -23,7 +23,7 @@ RELEASE_DIR ?= out
 export ACK_GINKGO_DEPRECATIONS := 1.16.5
 export ACK_GINKGO_RC=true
 
-PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+PROJECT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
```
rejoshed, line 26: When we were running CI, abspath wasn't present on the container.

Makefile
```diff
@@ -177,7 +177,6 @@ export KUBEBUILDER_ASSETS=$(PROJECT_DIR)/bin
 
 .PHONY: test
 test: generate-mocks lint bin/ginkgo bin/kubectl bin/kube-apiserver bin/etcd ## Run tests. At the moment this is only unit tests.
-	@./hack/testing_ginkgo_recover_statements.sh --add # Add ginkgo.GinkgoRecover() statements to controllers.
```
rejoshed, line 180: Test suite does this itself now.

api/v1beta1/cloudstackmachine_webhook_test.go
```diff
@@ -18,11 +18,12 @@ package v1beta1_test
 
 import (
 	"context"
+
 	"sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/test/dummies"
```
rejoshed, line 23: Guess my IDE likes to alphabetize.

pkg/cloud/cloud_suite_test.go
```diff
@@ -17,13 +17,76 @@ limitations under the License.
...
+	client          cloud.Client
+	realCSClient    *cloudstack.CloudStackClient
+	testDomainPath  string // Needed in before and in after suite.
 )
 
 func TestCloud(t *testing.T) {
 	RegisterFailHandler(Fail)
+	BeforeSuite(func() {
+		suiteConfig, _ := GinkgoConfiguration()
+		if !strings.Contains(suiteConfig.LabelFilter, "!integ") { // Skip if integ tests are filtered out.
```
rejoshed, line 44: If integ tests are not filtered out, then the test suite also creates a real client in a temporary account/domain that's wiped out at the end of testing.

test/helpers/client.go
```diff
@@ -0,0 +1,25 @@
+package helpers
+
```
rejoshed, line 2: The next few files are pretty much entirely about spinning up an account and domain for the duration of an integ test.

controllers/controllers_suite_test.go
```diff
@@ -183,6 +180,14 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = AfterSuite(func() {
+	projectDir := os.Getenv("PROJECT_DIR")
+	// Add ginkgo recover statements to controllers.
+	cmd := exec.Command(projectDir+"/hack/testing_ginkgo_recover_statements.sh", "--remove")
+	cmd.Stdout = os.Stdout
+	if err := cmd.Run(); err != nil {
+		fmt.Println("Refusing to run tests without ginkgo recover set.")
```
maxdrib, line 188: this error message doesn't make sense for cleanup

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.1.4
 	github.com/onsi/gomega v1.19.0
 	github.com/pkg/errors v0.9.1
+	github.com/smallfish/simpleyaml v0.1.0
```
maxdrib, line 13: what's the point of controllers/mocks/.keep?

pkg/cloud/cloud_suite_test.go
```diff
@@ -17,13 +17,76 @@ limitations under the License.
...
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/test/dummies"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/test/helpers"
+)
+
+var (
+	realCloudClient cloud.Client
+	client          cloud.Client
```
maxdrib, line 35: could you clarify the difference between these two? either a comment or renaming the variable. I'm not sure what `realCloudClient` means here either, that could be referring to anything in this day and age. Maybe mention that one is for the temporary test account user, and the other is for the credentials from the config file

pkg/cloud/isolated_network.go
```diff
@@ -247,8 +247,9 @@ func (c *client) GetOrCreateIsolatedNetwork(
 		if err = c.CreateIsolatedNetwork(zone, isoNet); err != nil {
 			return errors.Wrap(err, "creating a new isolated network")
 		}
+	} else {
+		isoNet.Spec.ID = net.ID
 	}
-	isoNet.Spec.ID = net.ID
```
maxdrib, line 251: why is this change being made? I think we want to set the ID if we're creating a new isolated network right?

test/helpers/client.go
```diff
@@ -0,0 +1,25 @@
...
+	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
+)
+
+func NewCSClient() (*cloudstack.CloudStackClient, error) {
+	projDir := os.Getenv("PROJECT_DIR")
+	conf := cloud.Config{}
+	ccPath := projDir + "/cloud-config"
+	if rawCfg, err := ini.Load(ccPath); err != nil {
+		return nil, errors.Wrapf(err, "reading config at path %s:", ccPath)
+	} else if g := rawCfg.Section("Global"); len(g.Keys()) == 0 {
```
maxdrib, line 18: can you use a constant for Global, or not use Global at all? We're taking a dependency on it even as we're planning to move away from using it. With a constant we'll at least know where it's being used when we're migrating. See https://github.com/aws/eks-anywhere/pull/2399/files for what the new schema will look like and how ini sections will be used as failure domain names

test/helpers/user.go
```diff
@@ -0,0 +1,190 @@
...
+	p.SetParentdomainid(parentID)
+	resp, err := csClient.Domain.CreateDomain(p)
+	if err != nil {
+		return "", err
+	}
+	return resp.Id, nil
+}
+
+// GetOrCreateDomain gets or creates a domain as specified in the passed domain object.
+func GetOrCreateDomain(csClient *cloudstack.CloudStackClient, domain *cloud.Domain) error {
```
maxdrib, line 61: is there any configuration on the cloudstack environment itself for this to work? Do domains/users have resource limits by default which need to be adjusted in anyway?

controllers/controllers_suite_test.go
rejoshed: Good point, thanks!

test/helpers/client.go
rejoshed: Max, I'm happy to make this a constant for now.

That said, I think the jury is still out on how to pass through the credentials when it comes to failure domains.

Myself and others think we should be moving forward with k8s secretes and a yaml file for credentials.

An INI requires an additional name field for a failure domain or a availability zone as is being referred to in the EKS-A design.

test/helpers/user.go
rejoshed: The tests all pass for me and I did not do any special configuration.

Could you please try running these tests to see if the new integ account/domain stuff works on your end? I think it should. Would just be `gingko -v ./pkg/...` so long as your setup matches the config in the e2e directory.

maxdrib: I got 48 tests passing with a bunch skipped like

```
S [SKIPPED] [0.001 seconds]
AffinityGroup Unit Tests
/Users/dribinm/workspaces/eksa/cluster-api-provider-cloudstack/pkg/cloud/affinity_groups_test.go:30
  AffinityGroup Integ Tests [BeforeEach]
  /Users/dribinm/workspaces/eksa/cluster-api-provider-cloudstack/pkg/cloud/affinity_groups_test.go:72
    Creates an affinity group.
    /Users/dribinm/workspaces/eksa/cluster-api-provider-cloudstack/pkg/cloud/affinity_groups_test.go:82

  Could not connect to ACS instance.
  In [BeforeEach] at: /Users/dribinm/workspaces/eksa/cluster-api-provider-cloudstack/pkg/cloud/affinity_groups_test.go:74
```
Any ideas how to resolve this? Some more logging would be helpful :)

pkg/cloud/cloud_suite_test.go
rejoshed: `cloud` is the cloud package being tested.

realCloudClient is a cloud.Client with a true connection to Apache CloudStack

Otherwise during testing we mock the connection to ACS.

I'll add some comments accordingly.

pkg/cloud/isolated_network.go
rejoshed: The following test fails without this change. 

https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/b9dc417e24dbd90e1c1eb5573d5c735444eca64d/pkg/cloud/isolated_network_test.go#L218

We were skipping this test prior.f

I'm going to add some comments, but basically there's a back and forth between the CloudStackIsolatedNetwork CRD/type here and the Network type.

I would like the two to be the same thing, but haven't been able to resolve that just yet.

In this case, if a CloudStackIsolatedNetwork is passed to this method without the ID set, and the Isolated network exists, then we will actually fail to set the ID of that network on the CloudStackIsolatedNetwork.

test/dummies/vars.go
```diff
@@ -202,8 +221,7 @@ func SetDummyCSMachineVars() {
 
 func SetDummyZoneVars() {
 	Zone1 = capcv1.Zone{Network: Net1}
-	Zone1.Name = "Zone1"
-	Zone1.ID = "FakeZone1ID"
+	Zone1.Name = GetYamlVal("CLOUDSTACK_ZONE_NAME")
```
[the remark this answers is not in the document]
rejoshed: What environment are you working against? Does Zone1 exist there?

If it does, then can we do a short call to explore this?

rejoshed:
@maxdrib I just fixed up that last skipped test. It's still slow even with the force stop. We don't run integs in the Pipeline atm anyway, so it is what it is I suppose. Please mark lgtm if you approve. Thanks!

PR #122, opened 2022-06-20
Fix intermittent e2e deploy test failures where empty string returned as the test html

jweite-amazon:
All clusters exhibiting this failure only do so for the first request made to the test httpd service.  Readiness probes did not help.

*Issue #, if available:*

*Description of changes:*
Added a delayed retry of the html fetch.

*Testing performed:*
Re-ran test until the problem occurred and was reported.  Observed that the delayed retry succeeded.

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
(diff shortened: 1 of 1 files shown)


test/e2e/common.go
```diff
@@ -187,9 +188,22 @@ func DeployAppToWorkloadClusterAndWaitForDeploymentReady(ctx context.Context, wo
...
+		return result, err
+	}
+	if result == "" {
+		// A single retry to accommodate occasional cases where an empty string is returned, ostensibly
+		//  because the service isn't fully ready.  Subsequent requests have always worked.
+		fmt.Println("Retrying html download")
+		time.Sleep(5 * time.Second)
+		runArgs[0] = "dummy2" // Assumed: container name is runArg zero.
+		result, err = KubectlExec(ctx, "run", workloadKubeconfigPath, runArgs...)
+	}
```
maxdrib, line 205: it seems like a for loop might be more robust than a single hardcoded retry, no?

jweite-amazon: I'm not really going for robust, just providing some resilience to slow startup.  If it doesn't succeed on the single retry I'm happy for the test to fail so we'll be inspired to figure out why.

test/e2e/common.go
```diff
@@ -187,9 +188,22 @@ func DeployAppToWorkloadClusterAndWaitForDeploymentReady(ctx context.Context, wo
...
+	if err != nil {
+		return result, err
+	}
+	if result == "" {
+		// A single retry to accommodate occasional cases where an empty string is returned, ostensibly
+		//  because the service isn't fully ready.  Subsequent requests have always worked.
+		fmt.Println("Retrying html download")
+		time.Sleep(5 * time.Second)
+		runArgs[0] = "dummy2" // Assumed: container name is runArg zero.
+		result, err = KubectlExec(ctx, "run", workloadKubeconfigPath, runArgs...)
```
maxdrib, line 204: should we return an error if result == "" again? It seems like with this implementation, no error will be returned

jweite-amazon: No, I don't consider that an error.  If the second attempt also returns "" with no error, then I'm interpreting this to mean that "" is the legitimate response.  The caller then gets to interpret that response as they see fit. In the deploy_app case, it'll fail the test for no-match against the expected return.

maxdrib (approved):

rohityadavcloud:
Appears basic build tests are okay
/lgtm
/approve

PR #124, opened 2022-06-21
Custom Metrics for ACS API calls. 

jweite-amazon:
One metric updater implemented in isolated_network.go.  Draft for feedback.

*Issue #, if available:*

*Description of changes:*
Added new metrics package defining custom prometheus metrics.
Added a CustomMetrics struct to the ACS client struct.
Modified a first ACS client call in pkg/cloud/isolated_network.go.

*Testing performed:*
Provisioned a cluster with isolated network, which is known to encounter a single "9999" error upon first call to AssociatePublicIPAddress.
Confirmed that a metrics port scrape includes the expected metric:
`# HELP acs_reconciliation_errors Count of reconciliation errors caused by ACS issues, bucketed by error code
# TYPE acs_reconciliation_errors counter
acs_reconciliation_errors{acs_error_code="9999"} 1
`


test/e2e/common.go
```diff
@@ -192,6 +192,37 @@ func DownloadFromAppInWorkloadCluster(ctx context.Context, workloadKubeconfigPat
...
 }
 
+func DownloadMetricsFromCAPCManager(ctx context.Context, bootstrapKubeconfigPath string) (string, error) {
+	// Expose the CAPC manager metrics port via a K8S service
+	runArgs := []string{
+		"--port=8080", "--target-port=metrics", "--name=capc-controller-manager-metrics", "--namespace=capc-system", "deployment", "capc-controller-manager",
+	}
+	_, err := KubectlExec(ctx, "expose", bootstrapKubeconfigPath, runArgs...)
+	if err != nil {
+		fmt.Println(err)
```
wongni, line 202: Remove if this is just for debugging

rejoshed: Either of
``` go
			Ω(KubectlExec(ctx, "expose", bootstrapKubeconfigPath, runArgs...)).Should(Succeed())
```
``` go
			_, err := KubectlExec(ctx, "expose", bootstrapKubeconfigPath, runArgs...)
			Ω(err).Error().ShouldNot(HaveOccurred())
```
Should work nicely if ginkgo/gomega is in use, and I was pretty sure it was in the E2Es.

rejoshed: Yeah, it's imported, so as long as it's setup right, you should be able to drop in either of the above snippets.

test/e2e/deploy_app.go
```diff
@@ -110,6 +110,14 @@ func DeployAppSpec(ctx context.Context, inputGetter func() CommonSpecInput) {
 
 		Expect(actualHtml).To(Equal(expectedHtml))
 
+		By("Confirming that the custom reconciliation error metric is scrape-able")
+		// BIG NOTE: The first reconciliation attempt of isolated_network!AssociatePublicIPAddress() returns
```
rejoshed, line 114: Ha, I like this.

**BIG NOTE**

test/e2e/common.go
jweite-amazon: I went with the second form.  In the first form Ω seemed to mix up the result and error return params.  Also, I needed the second form for the KubeCtl exec where I capture the response so I the func can return it.

wongni (approved):

pkg/cloud/affinity_groups.go
```diff
@@ -45,6 +45,7 @@ func (c *client) FetchAffinityGroup(group *AffinityGroup) (reterr error) {
 		affinityGroup, count, err := c.cs.AffinityGroup.GetAffinityGroupByID(group.ID)
 		if err != nil {
 			// handle via multierr
+			c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err)
```
rejoshed, line 48: Goodness. Seems we should be able to build something more universal than calling this on every potential ACS error.

You mentioned wanting to instrument the ACS-Go API, so I know you were thinking the same.

So, the ACS-Go API isn't easily instrumented, but is there maybe some way we can at least make these into one liners? Maybe a function that wraps an error and returns whatever the error was?

From there, we'd probably want to come up with a custom linter that enforces use of said function.

pkg/cloud/client.go
```diff
@@ -100,6 +104,6 @@ func (origC *client) NewClientFromSpec(cfg Config) (Client, error) {
 }
 
 func NewClientFromCSAPIClient(cs *cloudstack.CloudStackClient) Client {
-	c := &client{cs: cs, csAsync: cs}
+	c := &client{cs: cs, csAsync: cs, customMetrics: metrics.NewCustomMetrics()}
```
rejoshed, line 107: Will it be okay to recreate this `customMetrics:` object every time we reconcile? (performance and consistency wise?)

For `r.CSUser` we already create a new client for every reconcile call. But in that case it's created from an existing client, so we could make it reuse the existing `customMetrics:` object. 

Unfortunately when each reconcile request creates its own client it can't do that. So, we may want to initialize a single `customMetrics` object on the base reconciler and pass it to new clients as needed.

Again, this is only if a single consistent `customMetrics:` object is required or creation of said object is somewhat heavy.

pkg/cloud/isolated_network.go
```diff
@@ -44,6 +44,7 @@ type IsoNetworkIface interface {
 func (c *client) getOfferingID() (string, error) {
 	offeringID, count, retErr := c.cs.NetworkOffering.GetNetworkOfferingID(NetOffering)
 	if retErr != nil {
+		c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(retErr)
```
rejoshed, line 47: Yeah, I think we can add a function called `reportError` or something and instead of the above we can call:
``` Go
		return "", reportErr(retErr)
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// AcsCustomMetrics encapsulates all CloudStack custom metrics defined for the controller.
+type AcsCustomMetrics struct {
+	acsReconciliationErrorCount *prometheus.CounterVec
+	errorCodeRegexp             *regexp.Regexp
+}
+
+// NewCustomMetrics constructs an AcsCustomMetrics with all desired CloudStack custom metrics and any supporting resources.
+func NewCustomMetrics() AcsCustomMetrics {
```
rejoshed, line 33:
```suggestion
// NewCustomMetrics constructs an ACSCustomMetrics with all desired CloudStack custom metrics and any supporting resources.
func NewCustomMetrics() ACSCustomMetrics {
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+			customMetrics.acsReconciliationErrorCount = are.ExistingCollector.(*prometheus.CounterVec)
+		} else {
+			// Something else went wrong!
+			panic(err)
+		}
+	}
+
+	// ACS standard error messages of the form "CloudStack API error 431 (CSExceptionErrorCode: 9999):..."
+	//  This regexp is used to extract CSExceptionCodes from the message.
+	customMetrics.errorCodeRegexp, _ = regexp.Compile(".+CSExceptionErrorCode: ([0-9]+).+")
```
rejoshed, line 53: I ❤️ regexes.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+type AcsCustomMetrics struct {
+	acsReconciliationErrorCount *prometheus.CounterVec
+	errorCodeRegexp             *regexp.Regexp
+}
+
+// NewCustomMetrics constructs an AcsCustomMetrics with all desired CloudStack custom metrics and any supporting resources.
+func NewCustomMetrics() AcsCustomMetrics {
+	customMetrics := AcsCustomMetrics{}
+	customMetrics.acsReconciliationErrorCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
```
rejoshed, line 36: Here's where I was concerned about constantly making new `customMetrics:` objects.

If that's a new counter vector for every object, then won't the count binned by error code be reset on every client creation?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+			Name: "acs_reconciliation_errors",
+			Help: "Count of reconciliation errors caused by ACS issues, bucketed by error code",
+		},
+		[]string{"acs_error_code"},
+	)
+	if err := crtlmetrics.Registry.Register(customMetrics.acsReconciliationErrorCount); err != nil {
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
```
rejoshed, line 43: Maybe this registers the count with a central record?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	// ACS standard error messages of the form "CloudStack API error 431 (CSExceptionErrorCode: 9999):..."
+	//  This regexp is used to extract CSExceptionCodes from the message.
+	customMetrics.errorCodeRegexp, _ = regexp.Compile(".+CSExceptionErrorCode: ([0-9]+).+")
+
+	return customMetrics
+}
+
+// IncrementAcsReconciliationErrors accepts a CloudStack error message and increments the custom
```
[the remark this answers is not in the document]
rejoshed: I know we're not on this project much longer, but it'd be cool to add a linter that catches comment misnames. I'm sure I'm super guilty of this.

rejoshed (changes requested):
Please see comments. Thanks!

pkg/cloud/client.go
jweite-amazon: See my response to your comment about metrics.go line 35, below.

pkg/metrics/metrics.go
jweite-amazon: Yes, Register adds the new counter to the app's Prometheus registry.  The code here (and below) detects if this is the case, and if so returns that already registered one instead of a new one (the creation of which failed).

pkg/cloud/isolated_network.go
jweite-amazon: I'm not a fan of reportErr, because it doesn't really tell me what it's going to do. As a frequent code reader, I value explicitness over terseness.  I could have that function return the err that was passed in, so it could be placed in the return statement instead of on its own line.  I'm not feeling, though, like we really made a difference to readability/maintainability.  Same function call, just in a different line.

As the name EvaluateErrorAndIncrementAcsReconciliationErrorCounter implies, I sought get rid of *if err* conditional code.  (See [affinitygroups.go line 96](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/5efc0bc760cbf5df0181829b7f21d29bd11da69e/pkg/cloud/affinity_groups.go#L96)).

But our error handling patterns vary a bit.  Sometimes there's need for an else, so the if is still required.  Sometimes there's wrapping in Multierrors after the call.  

There were cases where I was tempted to re-structure the error handling, but ultimately decided the safe road of least-impact was the way to go.

After wrangling the variations, I concluded that this form pretty universally served all the situations without requiring variations, and went with it.

PR #132, opened 2022-07-02
Added e2e tests for upgrading clusters with insufficient compute resources

mrog:
*Issue #, if available:*

*Description of changes:*
Added e2e tests for upgrading an existing cluster, with insufficient compute resources for the upgrade

*Testing performed:*
Ran e2e tests


wongni (approved):

test/e2e/config/cloudstack.yaml
```diff
@@ -125,6 +126,7 @@ variables:
   CLOUDSTACK_INVALID_CONTROL_PLANE_MACHINE_OFFERING: "OfferingXXXX"
   CLOUDSTACK_EXTREMELY_LARGE_CONTROL_PLANE_MACHINE_OFFERING: "Extremely Large Instance"
```
pmotyka, line 127: Should the offering name be more descriptive of the compute resource being modeled?

davidjumani: Would it be better to name it "impossible offering" ? It is misleading unless the user reads the test details

test/e2e/config/cloudstack.yaml
```diff
@@ -127,6 +128,7 @@ variables:
...
   CLOUDSTACK_WORKER_MACHINE_OFFERING: "Medium Instance"
+  CLOUDSTACK_EXTREMELY_LARGE_WORKER_MACHINE_OFFERING: "Extremely Large Instance"
```
davidjumani, line 131: Here too

davidjumani:
Apart from that comments, looks good, tests pass

test/e2e/config/cloudstack.yaml
mrog: That's a good name for it.  I'll rename it.

mrog: That change is done and tested.

PR #139, opened 2022-07-22
[tests] Extend unit tests for cloud package

Pearl1594:
*Issue #, if available:*
Adds more unit tests for the cloud package

*Description of changes:*

*Testing performed:*


pkg/cloud/affinity_groups_test.go
```diff
@@ -48,22 +55,79 @@ var _ = Describe("AffinityGroup Unit Tests", func() {
 		mockCtrl.Finish()
 	})
 
-	It("fetches an affinity group", func() {
-		dummies.AffinityGroup.ID = "" // Force name fetching.
-		ags.EXPECT().GetAffinityGroupByName(dummies.AffinityGroup.Name).Return(&cloudstack.AffinityGroup{}, 1, nil)
+	Context("Fetch or Create Affinity group", func() {
```
rejoshed, line 58: Contexts are generally situational. 

https://onsi.github.io/ginkgo/#organizing-specs-with-container-nodes

You may want to use a describe here instead of a context.

``` Go
  Describe("Extracting the author's first and last name", func() {
    Context("When the author has both names", func() {
      It("can extract the author's last name", func() {   
```

Of course they all do the same thing though. : )

pkg/cloud/affinity_groups_test.go
```diff
@@ -48,22 +55,79 @@ var _ = Describe("AffinityGroup Unit Tests", func() {
...
+			ags.EXPECT().GetAffinityGroupByID(dummies.AffinityGroup.ID).Return(nil, -1, fakeError)
+			ags.EXPECT().NewCreateAffinityGroupParams(dummies.AffinityGroup.Name, dummies.AffinityGroup.Type).
+				Return(&cloudstack.CreateAffinityGroupParams{})
+			ags.EXPECT().CreateAffinityGroup(ParamMatch(And(NameEquals(dummies.AffinityGroup.Name)))).
+				Return(&cloudstack.CreateAffinityGroupResponse{}, nil)
+
+			Ω(client.GetOrCreateAffinityGroup(dummies.AffinityGroup)).Should(Succeed())
+		})
+
+		It("creates an affinity group if Name provided returns more than one affinity group", func() {
```
rejoshed, line 84: Now that I'm reading this, this behavior may not actually be what we want. This could be a good catch.

I think we probably want to return error if more than one affinity group comes back when listed by name?

We assumed the name to be unique.

rejoshed: At least within account/domain/endpoint the affinity group name is enforced as unique.
image

rejoshed: So I think we should never really see this happen, but I think we'd actually prefer an error here.

Either way, this test is still fine to merge. It specs the current behavior.

pkg/cloud/affinity_groups_test.go
```diff
@@ -48,22 +55,79 @@ var _ = Describe("AffinityGroup Unit Tests", func() {
...
+			dummies.AffinityGroup.ID = "" // Force name fetching.
+			agp := &cloudstack.CreateAffinityGroupParams{}
+			ags.EXPECT().GetAffinityGroupByName(dummies.AffinityGroup.Name).Return(nil, -1, fakeError)
+			ags.EXPECT().NewCreateAffinityGroupParams(gomock.Any(), gomock.Any()).Return(agp)
+			ags.EXPECT().CreateAffinityGroup(agp).Return(&cloudstack.CreateAffinityGroupResponse{}, nil)
+
+			Ω(client.GetOrCreateAffinityGroup(dummies.AffinityGroup)).Should(Succeed())
+		})
+
+		It("creates an affinity group if ID provided returns more than one affinity group", func() {
```
rejoshed, line 104: This also weirds me out.

If this test and the underlying implementation truly act this way, wouldn't this in the end return an error if working against a real ACS instance?

I would think it would then attempt to create the new Affinity group and fail due to the new name?

What do you think? If you agree this is odd, would you be up for taking on that change in implementation and the test case to make our intention here more clear?

My thinking is that this should return a nice descriptive error message that says we found more than one affinity group with that name.

Pearl1594: Ideally this is something that cannot happen, i.e., have 2 affinity groups with the same name - even via a DB hack, as there's a unique key constraint on the name column.

PR #142, opened 2022-08-01
api: Allow changes to CloudstackMachineTemplate

davidjumani:
*Description of changes:*

Allow changes in the following VM fields :
- Template : To allow Kubernetes version upgrades. If the user can not change these fields, they are stuck with the same version of Kubernetes
- Offering : To allow users to vertically scale the cluster

We can also discuss the feasibility of changing other fields such as the ssh key, disk offering, etc. as users might want to change these as well while upgrading a cluster

*Testing performed:*
```
------------------------------
CloudStackMachineTemplate webhook When updating a CloudStackMachineTemplate
  should allow VM template updates to the CloudStackMachineTemplate
  /home/djumani/lab/capi/api/v1beta1/cloudstackmachinetemplate_webhook_test.go:76
•
------------------------------
CloudStackMachineTemplate webhook When updating a CloudStackMachineTemplate
  should allow VM offering updates to the CloudStackMachineTemplate
  /home/djumani/lab/capi/api/v1beta1/cloudstackmachinetemplate_webhook_test.go:89
•
```

### DIFF
--- a/api/v1beta1/conversion_test.go
+++ b/api/v1beta1/conversion_test.go
@@ -26,8 +26,6 @@ import (
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
-
-
 var _ = Describe("Conversion", func() {
 	BeforeEach(func() { // Reset test vars to initial state.
 	})
--- a/api/v1beta2/cloudstackmachinetemplate_webhook.go
+++ b/api/v1beta2/cloudstackmachinetemplate_webhook.go
@@ -17,11 +17,8 @@ limitations under the License.
 package v1beta2
 
 import (
-	"fmt"
-	"reflect"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/webhookutil"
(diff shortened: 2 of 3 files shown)


api/v1beta1/cloudstackmachine_webhook.go
```diff
@@ -84,15 +84,13 @@ func (r *CloudStackMachine) ValidateUpdate(old runtime.Object) error {
 	}
 	oldSpec := oldMachine.Spec
 
-	errorList = webhookutil.EnsureBothFieldsAreEqual(r.Spec.Offering.ID, r.Spec.Offering.Name, oldSpec.Offering.ID, oldSpec.Offering.Name, "offering", errorList)
```
maxdrib, line 87: I think the idea was for cloudstackmachines to be immutable - if a user wants to upgrade, they can create a new one and remove the old one. EKS-A supports these upgrades so users should not be stuck on a particular version or offering.

maxdrib: @rejoshed does that sound right?

davidjumani: Right, I'll update it to only allow changes to the MachineTemplate

davidjumani:
@maxdrib @rejoshed @jweite-amazon I'd also suggest allowing users to modify other fields such as ssh key, details, etc but only during a template / offering change (an upgrade) as users might want to change them during an upgrade. Thoughts ?

jweite-amazon:
I'm good with this @davidjumani and will approve after rebase.

> modify other fields such as ssh key, details, etc but only during a template / offering change (an upgrade) as users might want to change them during an upgrade. Thoughts ?

I'm good with changes to other fields in the template too.  What's the reason for making them mutable only as part of a template/offering change?  I can imagine, for instance, wanting to change an ssh key without wanting to change template/offering (and will tolerate the machine replacement process to do so).

rejoshed:
@jweite-amazon @maxdrib @davidjumani 

My understanding was that a change of any of these fields was to be done by pointing a MachineDeployment or KubeACMControlPlane to a new machine template reference.

In this manner, any field can be modified.

This triggers a rolling update.

Otherwise, machines that already exist won't update, and we wouldn't be honoring the declarative nature of our cluster specification.

Also, sorry for the late reply.

davidjumani:
> My understanding was that a change of any of these fields was to be done by pointing a MachineDeployment or KubeACMControlPlane to a new machine template reference.

This might be one way, but a user can simply modify the existing MachineTemplate in the cluster-spec.yaml to do the same thing, which is much more intuitive, especially for newbies. This way too triggers a rolling upgrade and IMO this is much simpler than creating a new template and pointing to it

davidjumani:
@rejoshed Can you pl have a look

rejoshed:
@davidjumani  I'll check into the failing test sometime in the next day or two.

Looks like the `FailureDomainController` is asking for resolution of the ZoneId from cloudstack, but we've not mocked that call.

I don't really understand why it would be asking for it now, and not without your changes.

> Can you pl have a look

rejoshed:
Kicked off a rerun of the tests and they passed. Ran locally too. Merging.

PR #143, opened 2022-08-02
Update machine state checker so it doesn't delete machines when it shouldn't

mrog:
Updating machine state checker so it won't delete a machine because of the CAPI phase, unless the phase is Provisioned.  (It previously looked for any phase other than Running.)

*Issue #, if available:*

*Description of changes:*

*Testing performed:*
Manually tested with cluster creation.

### DIFF
--- a/controllers/cloudstackmachinestatechecker_controller.go
+++ b/controllers/cloudstackmachinestatechecker_controller.go
@@ -79,23 +79,57 @@ func (r *CloudStackMachineStateCheckerReconciliationRunner) Reconcile() (ctrl.Re
 		return r.ReturnWrappedError(err, "failed to resolve VM instance details")
 	}
 
-	csRunning := r.CSMachine.Status.InstanceState == "Running"
+	const requeueAfter = 5 * time.Second
+
+	csState := r.CSMachine.Status.InstanceState
 	csTimeInState := r.CSMachine.Status.TimeSinceLastStateChange()
+
+	if csState == "" || csTimeInState < 0 {
+		// Wasn't able to get the CS state, and can't continue without it.
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
+
 	capiPhase := capiv1.MachinePhase(r.CAPIMachine.Status.Phase)
 	capiRunning := capiPhase == capiv1.MachinePhaseRunning
-	deleteInvoked := capiPhase == capiv1.MachinePhaseDeleting || capiPhase == capiv1.MachinePhaseDeleted
 
-	// capiTimeout indicates that a new VM is running, but it isn't reachable due to a network issue or a misconfiguration.
-	// When this happens, the machine should be deleted or else the cluster won't ever recover.
-	capiTimeout := csRunning && !capiRunning && csTimeInState > 5*time.Minute
+	if capiPhase == capiv1.MachinePhaseDeleting || capiPhase == capiv1.MachinePhaseDeleted {
+		// The machine is being deleted, so do nothing for now.
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
+
+	var (
+		csRunning     bool
+		csStopTimeout bool
+		csOtherState  bool
+		capiTimeout   bool
+	)
+
+	switch csState {
+	case "Running":
+		csRunning = true
+		// capiTimeout indicates that a new VM is running, but it isn't reachable due to a network issue or a
+		// misconfiguration.  When this happens, the machine should be deleted or the cluster won't recover.
+		capiTimeout = !capiRunning && csTimeInState > 5*time.Minute
+	case "Starting":
+	case "Migrating":
+		// Give CS time to do what it needs to do
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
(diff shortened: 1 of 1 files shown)


controllers/cloudstackmachinestatechecker_controller.go
```diff
@@ -82,10 +82,11 @@ func (r *CloudStackMachineStateCheckerReconciliationRunner) Reconcile() (ctrl.Re
 	csRunning := r.CSMachine.Status.InstanceState == "Running"
 	csTimeInState := r.CSMachine.Status.TimeSinceLastStateChange()
 	capiRunning := r.CAPIMachine.Status.Phase == "Running"
+	capiProvisioned := r.CAPIMachine.Status.Phase == "Provisioned"
 
 	// capiTimeout indicates that a new VM is running, but it isn't reachable due to a network issue or a misconfiguration.
 	// When this happens, the machine should be deleted or else the cluster won't ever recover.
-	capiTimeout := csRunning && !capiRunning && csTimeInState > 5*time.Minute
+	capiTimeout := csRunning && capiProvisioned && csTimeInState > 5*time.Minute
```
maxdrib, line 89: Could you elaborate more on the change in scope? How will this change help us address the issues we've observed? 

Also, how was this change tested?

mrog: I just a new version a few minutes ago that'll work better.  The main difference from the old behavior is that it won't repeatedly try to delete the same machine.  If it sees that it's already being deleted, it does nothing.  This prevents clutter in the log, and avoids some wasted API calls.

I tested it by creating a new cluster and disabling the VM network interfaces as the machines were starting.  This prevented them from joining the cluster, while CS still showed the status as running.  After the 5 minute waiting period, each disabled VM was deleted exactly once.  I tested with both CP and worker machines.

mrog:
@rejoshed I made some changes based on our conversation yesterday.

controllers/cloudstackmachinestatechecker_controller.go
```diff
@@ -81,21 +81,23 @@ func (r *CloudStackMachineStateCheckerReconciliationRunner) Reconcile() (ctrl.Re
...
+	} else if (!csRunning || capiTimeout) && !deleteInvoked {
 		r.Log.Info("CloudStack instance in bad state",
 			"name", r.CSMachine.Name,
 			"instance-id", r.CSMachine.Spec.InstanceID,
 			"cs-state", r.CSMachine.Status.InstanceState,
 			"cs-time-in-state", csTimeInState.String(),
-			"capi-phase", r.CAPIMachine.Status.Phase)
+			"capi-phase", capiPhase)
 
 		if err := r.K8sClient.Delete(r.RequestCtx, r.CAPIMachine); err != nil {
```
maxdrib, line 102: are we still sure that deleting the CAPI machine is the right thing to do? It feels a bit risky and like a shortcut. Have we considered deleting the CloudStackMachine instead, or is there something preventing us from doing that?

jweite-amazon: FWIW, the CAPI-provided Machine Health Check mechanisms deletes CAPI machines.  Its subordinate InfrastructureMachine (i.e., CloudStackMachine) gets deleted with it.

jweite-amazon: [PR-58](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/58) actually addressed a problem caused when we delete CloudStackMachine instead of Machine.

jweite-amazon: /approved

controllers/cloudstackmachinestatechecker_controller.go
```diff
@@ -81,21 +81,23 @@ func (r *CloudStackMachineStateCheckerReconciliationRunner) Reconcile() (ctrl.Re
 
 	csRunning := r.CSMachine.Status.InstanceState == "Running"
 	csTimeInState := r.CSMachine.Status.TimeSinceLastStateChange()
-	capiRunning := r.CAPIMachine.Status.Phase == "Running"
+	capiPhase := capiv1.MachinePhase(r.CAPIMachine.Status.Phase)
+	capiRunning := capiPhase == capiv1.MachinePhaseRunning
+	deleteInvoked := capiPhase == capiv1.MachinePhaseDeleting || capiPhase == capiv1.MachinePhaseDeleted
```
maxdrib, line 86: I thought the main issue we were trying to address was deleting CAPI machines on move operations in a slow Cloudstack environment? I don't see that logic being implemented in this PR. Do we have a plan to address it in a follow-up PR?

mrog: That's an unrelated issue to be addresses separately.

mrog: That's an unrelated issue to be addresses separately.

mrog:
I made some more changes to handle two situations: VM migration, and missing CS status

mrog:
We decided to disable machine state checker until we figure out why control plane machines don't always get replaced.  I'm closing this PR for now.

PR #144, opened 2022-08-02
Implement client caching

wongni:
*Issue #, if available:*

*Description of changes:*
Instead of creating a client per reconciling, use a client cache to reduce the number of client connections

*Testing performed:*
Created a cluster with 1 CP and 10 workers across two ACS endpoints. With debug output, only two clients were created and the cache clients were used per reconciling


wongni:
Created https://github.com/rejoshed/cluster-api-provider-cloudstack/pull/1

PR #145, opened 2022-08-09
Use listall=true when resolving network to support network created under sub-domains

wongni:
*Issue #, if available:*

*Description of changes:*
When the specified shared network is created under a sub-domain resolving network fails to get the network details of the network so a new isolated network is created. This PR adds listall=true when resolving networks to support those networks created under sub-domains

*Testing performed:*
Manual e2e testing


wongni:
I tested the issue reported by the customer but I wasn't able to reproduce it in the multi-endpoint support branch because the zone controller now uses the client created for the sub-domain when resolving network of the zone. 

```
	if err := r.CSUser.ResolveNetworkForZone(r.ReconciliationSubject); err != nil &&
		!csCtrlrUtils.ContainsNoMatchSubstring(err) {
		return ctrl.Result{}, errors.Wrap(err, "resolving Cloudstack network information")
	}
```

Issue #113, opened 2022-06-03
Inconsistent JSON tags


davidjumani:
In certain structs, the JSON tags case is not consistent. 
Eg: `CloudStackMachineSpec.AffinityGroupIDs` is tagged to `affinitygroupids` while `CloudStackMachineSpec.ZoneID` is tagged to `zoneID`. 

There are others and it would be great to standardise the tags to a certain convention prior to an initial release so that down the line there won't be any compatibility issues

rejoshed:
Please see linked PR. I only found a few JSON tags that I could tell needed changing.

Hopefully we can close this issue with said PR.

PR #147, opened 2022-08-10
Feature/hash failure domain naming

rejoshed:
*Issue #, if available:*

*Description of changes:*
Meta naming for failure domain custom resource objects currently follows `failureDomainName` + `-` + `clusterName`.

This was to work around EKS-A's decision to put all clusters in a single namespace.

Unfortunately k8s object names can only be 64 characters long and this leads to the concern that CAPC will break for any combination of names that leads to >64 character failure domain naming.

This uses a hash of the failure domain name and the cluster name to predictably generate a failure domain's meta name.

Also should close https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/113#issue-1259386590

*Testing performed:*
All unit tests passing locally.

Ran e2e PR-Blocking tests successfully.


api/v1beta2/cloudstackmachine_types.go
```diff
@@ -63,7 +63,7 @@ type CloudStackMachineSpec struct {
 
 	// Optional affinitygroupids for deployVirtualMachine
 	// +optional
-	AffinityGroupIDs []string `json:"affinitygroupids,omitempty"`
+	AffinityGroupIDs []string `json:"affinityGroupIDs,omitempty"`
```
rejoshed, line 66: Just a fixup for JSON tagging consistency.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -84,17 +84,14 @@ func (r *CloudStackFailureDomainReconciliationRunner) Reconcile() (retRes ctrl.R
...
-	}
 
 	// Check if the passed network was an isolated network or the network was missing. In either case, create a
 	// CloudStackIsolatedNetwork to manage the many intricacies and wait until CloudStackIsolatedNetwork is ready.
 	if r.ReconciliationSubject.Spec.Zone.Network.ID == "" ||
 		r.ReconciliationSubject.Spec.Zone.Network.Type == infrav1.NetworkTypeIsolated {
 		netName := r.ReconciliationSubject.Spec.Zone.Network.Name
-		if res, err := r.GenerateIsolatedNetwork(netName)(); r.ShouldReturn(res, err) {
+		if res, err := r.GenerateIsolatedNetwork(
+			netName, func() string { return r.ReconciliationSubject.Spec.Name })(); r.ShouldReturn(res, err) {
```
rejoshed, line 94: Need to pass failure domain name in here.

config/manager/manager.yaml
```diff
@@ -15,8 +15,8 @@ spec:
       labels:
         control-plane: capc-controller-manager
     spec:
-      securityContext:
-        runAsNonRoot: true
+      # securityContext:
```
wongni, line 18: Please remote comments

test/e2e/config/cloudstack.yaml
```diff
@@ -113,7 +113,7 @@ variables:
   NODE_DRAIN_TIMEOUT: "60s"
 
   CLOUDSTACK_FD1_NAME: "fd1"
-  CLOUDSTACK_FD1_SECRET_NAME: "secret1"
+  CLOUDSTACK_FD1_SECRET_NAME: "myendpoint"
```
wongni, line 116: Please revert this back to secret1

api/v1beta2/cloudstackmachine_types.go
maxdrib, line 66: I don't think we should change this now. EKS-A is already and other customers may also be using this field and changing the serialization/deserialization patterns feels risky, with no gain. I'm inclined to keep these as they are and fix later on in an intentional way

Unless this is strictly for the v1beta2 types, and not v1beta1? Same for the other changes involving json

wongni (approved):

controllers/utils/failuredomains.go
```diff
@@ -73,22 +71,30 @@ func (r *ReconciliationRunner) GetFailureDomains(fds *infrav1.CloudStackFailureD
 	}
 }
 
+// GetFailureDomainByName gets a singel FailureDomain by name and requeues if it's not found.
```
maxdrib, line 74:
```suggestion
// GetFailureDomainByName gets a single FailureDomain by name and requeues if it's not found.
```

api/v1beta2/cloudstackisolatednetwork_types.go
```diff
@@ -41,7 +41,7 @@ type CloudStackIsolatedNetworkSpec struct {
 
 	//+k8s:conversion-gen=false
 	// FailureDomain -- the FailureDomain the network is placed in.
```
maxdrib, line 43: update comment

maxdrib (commented):
Why are all these test data files deleted?

wongni:
> Why are all these test data files deleted?

They are generated files via `kustomize` during executing run-e2e make target

api/v1beta2/cloudstackmachine_types.go
rejoshed: Indeed, this is only on v1beta2 types.

That's why I wanted to jamb this into this release. It won't be easy to do outside of an API version bump.

api/v1beta2/cloudstackisolatednetwork_types.go
rejoshed: Name changed.

controllers/utils/failuredomains.go
rejoshed: Changed.

maxdrib (approved):

PR #148, opened 2022-08-12
Addressing bug where account/domain was lost

maxdrib:
*Issue #, if available:*

*Description of changes:*
While upgrading from EKS-A v0.10.1 to the latest version (with latest CAPC), I observed that the account/domain information vanished from the CloudStackCluster spec. I suspect that as part of the EKS-A upgrade, we somehow end up performing a conversion from v1beta2 to v1beta1 and lose this context. 

*Testing performed:*
Upgrade did not cause account/domain information to be lost.

### DIFF
--- a/api/v1beta1/conversion.go
+++ b/api/v1beta1/conversion.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"context"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	conv "k8s.io/apimachinery/pkg/conversion"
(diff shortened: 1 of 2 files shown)


api/v1beta1/conversion.go
```diff
@@ -51,6 +51,8 @@ func Convert_v1beta1_CloudStackCluster_To_v1beta2_CloudStackCluster(in *CloudSta
 func Convert_v1beta2_CloudStackCluster_To_v1beta1_CloudStackCluster(in *v1beta2.CloudStackCluster, out *CloudStackCluster, scope conv.Scope) error {
 	out.ObjectMeta = in.ObjectMeta
 	out.Spec = CloudStackClusterSpec{
+		Account:              in.Spec.FailureDomains[0].Account,
```
wongni, line 54: Don't we have to check the size of in.Spec.FailureDomains?

rejoshed (approved):

api/v1beta1/conversion.go
```diff
@@ -49,8 +50,13 @@ func Convert_v1beta1_CloudStackCluster_To_v1beta2_CloudStackCluster(in *CloudSta
 
 //nolint:golint,revive,stylecheck
 func Convert_v1beta2_CloudStackCluster_To_v1beta1_CloudStackCluster(in *v1beta2.CloudStackCluster, out *CloudStackCluster, scope conv.Scope) error {
+	if len(in.Spec.FailureDomains) < 1 {
+		return fmt.Errorf("v1beta2 to v1beta1 conversion not supported when ")
```
wongni:
```suggestion
		return fmt.Errorf("v1beta2 to v1beta1 conversion not supported when FailureDomains are empty")
```

wongni (changes requested):

api/v1beta1/conversion.go
maxdrib: Fixed

wongni (approved):

PR #151, opened 2022-08-17
customize v1beta1 affinity group conversion to v1beta2

wanyufe:
*Issue #, if available:*
Add default failure domain name when converting v1beta1 affinity group to v1beta2 affinity group. This is only needed when upgrade 
*Description of changes:*
Assign first failure domain name to v1beta1 affinity groups since the first failure domain is the previous version's only failure domain.

Commented out cloudstackZone and cloudstackfailuredomain patch since cloudstackZone does not exist in v1beta2 and cloudstackFailuredomain does not exist in v1beta1
*Testing performed:*
Manually tested. 

### DIFF
--- a/api/v1beta1/cloudstackaffinitygroup_conversion.go
+++ b/api/v1beta1/cloudstackaffinitygroup_conversion.go
@@ -17,7 +17,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
+	conv "k8s.io/apimachinery/pkg/conversion"
 	"sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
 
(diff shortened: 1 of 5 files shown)


api/v1beta1/cloudstackaffinitygroup_types.go
```diff
@@ -46,7 +48,7 @@ type CloudStackAffinityGroupStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
-
+//+k8s:conversion-gen=false
```
maxdrib, line 51: don't we want conversion on affinity group?

wongni (approved):

api/v1beta1/cloudstackaffinitygroup_conversion.go
```diff
@@ -30,3 +33,44 @@ func (dst *CloudStackAffinityGroup) ConvertFrom(srcRaw conversion.Hub) error { /
...
+	key := client.ObjectKey{Name: clusterName, Namespace: namespace}
+	if err := v1beta2.K8sClient.Get(context.TODO(), key, csCluster); err != nil {
+		return err
+	}
+
+	out.Spec = v1beta2.CloudStackAffinityGroupSpec{
+		Type:              in.Spec.Type,
+		Name:              in.Spec.Name,
+		ID:                in.Spec.ID,
+		FailureDomainName: csCluster.Spec.FailureDomains[0].Name,
```
maxdrib, line 53: we might want to check for len(failuredomains) before making this change?

also, what happens if this conversion happens before the csCluster is converted?

maxdrib: lastly, why are we hardcoding this affinity group to always be associated with the first failuredomain? Couldn't that be wrong, even for our current upgrade case?

rejoshed: > why are we hardcoding this affinity group to always be associated with the first failuredomain? Couldn't that be wrong, even for our current upgrade case?

I'm not sure how it could be wrong. The unique aspects of a failure domain are the account/domain, zone, and endpoint secret to reference.

Affinity groups do not consider zones.

The account/domain and endpoint secret willl be the same across failure domains on upgrade.

rejoshed: > we might want to check for len(failuredomains) before making this change?

At least one failure domain should be present or the verification webhook would deny it. I personally prefer not to recheck when we have other assumptions about it, but I don't feel strongly either way.

rejoshed: > also, what happens if this conversion happens before the csCluster is converted? I see we're using the v1beta2 client - does that mean failuredomains will always be set? My concern is about getting a v1beta1 csCluster here and throwing a panic

As far as I'm aware, when reading the v1beta1 version of the cluster from ETCD, it will be converted first and other than conversion methods specifically targeting a type, we shouldn't get a v1beta1 type from a get call.

This does introduce the idea of a circular dependency on updated types, but I don't think we have one here.

api/v1beta1/cloudstackaffinitygroup_conversion.go
```diff
@@ -30,3 +33,44 @@ func (dst *CloudStackAffinityGroup) ConvertFrom(srcRaw conversion.Hub) error { /
 	src := srcRaw.(*v1beta2.CloudStackAffinityGroup)
 	return Convert_v1beta2_CloudStackAffinityGroup_To_v1beta1_CloudStackAffinityGroup(src, dst, nil)
 }
+
+//nolint:golint,revive,stylecheck
```
rejoshed, line 37: Why are we skipping linting here? I'm probably fine with it, I'd just like to understand.

api/v1beta1/cloudstackaffinitygroup_conversion.go
```diff
@@ -30,3 +33,44 @@ func (dst *CloudStackAffinityGroup) ConvertFrom(srcRaw conversion.Hub) error { /
...
+	}
+
+	out.Status = v1beta2.CloudStackAffinityGroupStatus{
+		Ready: in.Status.Ready,
+	}
+	return nil
+}
+
+//nolint:golint,revive,stylecheck
+func Convert_v1beta2_CloudStackAffinityGroup_To_v1beta1_CloudStackAffinityGroup(in *v1beta2.CloudStackAffinityGroup, out *CloudStackAffinityGroup, s conv.Scope) error {
```
rejoshed, line 63: It's interesting that we have these backward conversion methods, but they're only partially functional.

I wonder if we might want to go ahead with annotations as a way to make them fully functional?

api/v1beta1/cloudstackaffinitygroup_types.go
rejoshed: This is auto-generation of conversion methods. Yufei is hand writing the conversion, so we don't want conversion-gen to write these.

rejoshed (commented):
I see you've manually tested per your PR description, but what exactly was manually tested?

Did you create a cluster and upgrade it?

Were the old affinity groups deleted? Otherwise generally looks good to me aside from my small nit questions.

api/v1beta1/cloudstackaffinitygroup_conversion.go
jweite-amazon: > I'm not sure how it could be wrong.
@rejoshed, could you elaborate further?  If a cluster has multiple FDs defined for it, why does it make sense to arbitrarily pick the first one to associate the AG with?

rejoshed: In our upgrade scenario there was only ever one of each of the details that matters to an affinity group.

So, when upgrading without modifications (the limitation we agreed was reasonable) it doesn't matter to an affinity group which failure domain is picked.

And in this case we immediately set this custom resource object to be deleted, so it goes away before more failure domains can be introduced. This is because they would be introduced in a secondary cluster modification. Upgrade and the introduction of additional failure domains are a two step process. 1 to upgrade v1beta1 to v1beta2 and another to add or remove failure domains pointing to additional ACS management host endpoints.

PR #156, opened 2022-08-26
Update failuredomain remove logic

wanyufe:
*Issue #, if available:*
Currently VM in deleted failure domain are simply deleted and depends reconciler to recover them back.
*Description of changes:*

1. if etcd VM, needs to trigger etcdadmcluster rollout
2. if CP VM, and no etcd VM, need kubeadmcontrolplane rollout
3. if no CP/etcd, only worker VM, needs machine deployment roll out

*Testing performed:*
Manual testing.


controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+}
+
+func triggerEtcdClusterRollout(machines []infrav1.CloudStackMachine, r *CloudStackFailureDomainReconciliationRunner) (ctrl.Result, error) {
+	var etcdMachine infrav1.CloudStackMachine
+	etcdMachineFound := false
+	for _, machine := range machines {
+		for _, ref := range machine.OwnerReferences {
+			if ref.Kind == "EtcdadmCluster" {
+				etcdMachineFound = true
+				etcdMachine = machine
```
maxdrib, line 178: why do you need to save the machine? Are you trying to get the etcdOwnerRef? Why not just grab that in here instead of persisting the etcdMachine, then looping through its ownerRefs later on?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+		return result, err
+	}
+
+	result, err = triggerMachineDeploymentRollout(machines.Items, r)
+	if err != nil {
+		return result, err
+	}
+
+	if len(machines.Items) > 0 {
+		return r.RequeueWithMessage("FailureDomain still has machine(s) in it.")
```
maxdrib, line 166: I think this message could use more detail, in isolation of other context. Would it be more accurate/meaningful to say "Deleted FailureDomain still has machine(s) in it"?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+	for _, machine := range machines {
+		for _, ref := range machine.OwnerReferences {
+			if ref.Kind == "EtcdadmCluster" {
+				etcdMachineFound = true
+				etcdMachine = machine
+			}
+		}
+	}
+	if !etcdMachineFound {
+		r.Log.Info("Clear machine: no etcd VM found.")
```
maxdrib, line 183: I feel like these logs should be much more meaningful and provide independent context into what the code is doing. `no etcd VM found` doesn't mean anything to me. It would help to explain why we were looking for etcd vm's and the implication of not finding any

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+		return ctrl.Result{}, nil
+	}
+	for _, ref := range etcdMachine.OwnerReferences {
```
maxdrib, line 186: would an etcd machine ever have multiple ownerRefs?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+				func() error {
+					return unstructured.SetNestedField(
+						etcdadmCluster.Object,
+						csMachineTemplate.Name,
+						"spec", "infrastructureTemplate", "name",
+					)
+				}); err != nil {
+				return ctrl.Result{}, err
+			}
+			r.Log.Info(fmt.Sprintf("Clear machine: etcdadmcluster infrastructureTemplate cloudstackmachinetemplate name %s patched", csMachineTemplate.Name))
```
maxdrib, line 231: could you move this into a helper function to make it more readable? I also wonder if we could reuse some of this code between the different deployment triggers

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+			r.Log.Info("Clear machine: retrieved etcdadmcluster.")
+			csMachineTemplateName, ok := etcdMachine.Annotations["cluster.x-k8s.io/cloned-from-name"]
+			if !ok {
+				result, err := r.RequeueWithMessage("annotations['cluster.x-k8s.io/cloned-from-name'] not found.", "cloudstackmachine", etcdMachine.Name)
+				return result, err
+			}
+			r.Log.Info(fmt.Sprintf("Clear machine: machine template name %s decided", csMachineTemplateName))
+			csMachineTemplateNameInEtcdadmCluster, found, err := unstructured.NestedString(etcdadmCluster.Object, "spec", "infrastructureTemplate", "name")
+			if err != nil || !found {
+				return ctrl.Result{}, errors.New("etcdadmcluster spec.infrastructureTemplate.name not found or not string")
```
maxdrib, line 203: maybe also log the etcdadmcluster name for debugging ease? It might also help to return the original error for context

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+			}
+			if csMachineTemplateName != csMachineTemplateNameInEtcdadmCluster {
+				return ctrl.Result{}, errors.Errorf("cloudstackmachinetemplate in machine %s and etcdadmcluster %s are different", etcdMachine.Name, ref.Name)
```
maxdrib, line 206: it's weird that we're only looking at one machine. Shouldn't we look at all the etcd machines?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+			}
+			csMachineTemplate := &infrav1.CloudStackMachineTemplate{}
+			if err := r.K8sClient.Get(r.RequestCtx, client.ObjectKey{Namespace: etcdMachine.Namespace, Name: csMachineTemplateName}, csMachineTemplate); err != nil {
+				return ctrl.Result{}, err
+			}
+			r.Log.Info(fmt.Sprintf("Clear machine: machine template name %s retrieved", csMachineTemplateName))
+			csMachineTemplate.Name = fmt.Sprintf("%s-template-%d", ref.Name, time.Now().UnixNano()/int64(time.Millisecond))
+			csMachineTemplate.SetAnnotations(map[string]string{})
```
maxdrib, line 214: can you explain how you're triggering the rollout? It seems like you're creating a new cloudstackmachinetemplate and then setting that on the etcdadmcluster. But won't that mess up EKS-A's controller, which is generating and applying its own CloudStackMachineTemplates and EtcdAdmCluster resources?

maxdrib (changes requested):
There's a bunch of new logic here. How can we cover these cases with automated unit tests?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,222 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
 	if err := r.K8sClient.List(r.RequestCtx, machines, client.MatchingLabels{infrav1.FailureDomainLabelName: r.ReconciliationSubject.Name}); err != nil {
 		return ctrl.Result{}, err
 	}
-	// Deleted CAPI machines for CloudStack machines found.
-	for _, machine := range machines.Items {
+	r.Log.Info(fmt.Sprintf("Clear machine: %d found.", len(machines.Items)))
```
maxdrib, line 143: do we need to add a taint to the failure domain when it's scheduled for deletion to prevent new machines from being assigned to it?

controllers/cloudstackfailuredomain_controller.go
maxdrib: never mind, seems like it can ref to a CAPI machine and an etcd cluster

controllers/cloudstackfailuredomain_controller.go
wanyufe: To trigger an etcdadmcluster rollout, all needed info can be obtained from one etcd cloudstack machine.

controllers/cloudstackfailuredomain_controller.go
wanyufe: `return r.RequeueWithMessage("failuredomain delete: FailureDomain still has machine(s) in it.", "failuredomain", r.ReconciliationSubject.Spec.Name)`

controllers/cloudstackfailuredomain_controller.go
wanyufe: I combined two loop into one. And also grabbed the ref.

controllers/cloudstackfailuredomain_controller.go
wanyufe: those logs are for debugging purpose, I removed it.

controllers/cloudstackfailuredomain_controller.go
maxdrib: what about these machines makes them noteworthy? I would say 
```
return r.RequeueWithMessage("Deleted FailureDomain still has machine(s) in it which need to be cleared. Requeueing", "failuredomain", r.ReconciliationSubject.Spec.Name)
```

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,215 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+					timeNowStr := time.Now().Format(time.RFC3339)
+					md.Spec.Template.Annotations["cluster.x-k8s.io/restartedAt"] = timeNowStr
+					patcher, err := patch.NewHelper(md, r.K8sClient)
+					if err != nil {
+						return ctrl.Result{}, err
+					}
+					if err := patcher.Patch(r.RequestCtx, md); err != nil {
+						return ctrl.Result{}, err
+					}
+					r.Log.Info(fmt.Sprintf("failuredomain delete: machine deployment name %s add annotation restartedAt %s", mdName, timeNowStr))
```
maxdrib, line 269: how does this work when a cluster has multiple worker node groups? Has that been tested at all?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,215 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+	for _, machine := range machines {
+		for _, ref := range machine.OwnerReferences {
+			if ref.Kind == "KubeadmControlPlane" {
+				cpMachineFound = true
+				cpMachine = machine
+				ownerRef = ref
+			}
+			// etcdadmcluster rollout will trigger control plane rollout automatically.
+			// if etcd VM exists, trigger etcdadmcluster rollout is enough, no need to trigger control plane VM rollout.
+			if ref.Kind == "EtcdadmCluster" {
```
maxdrib, line 296: isn't there a constant for this value somewhere that we can use?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,215 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+	return ctrl.Result{}, nil
+}
+
+func triggerControlPlaneRollout(machines []infrav1.CloudStackMachine, r *CloudStackFailureDomainReconciliationRunner) (ctrl.Result, error) {
+	var cpMachine infrav1.CloudStackMachine
+	var ownerRef metav1.OwnerReference
+	cpMachineFound := false
+	for _, machine := range machines {
+		for _, ref := range machine.OwnerReferences {
+			if ref.Kind == "KubeadmControlPlane" {
```
maxdrib, line 289: this should be a constant

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,215 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+		r.Log.Info(fmt.Sprintf("failuredomain delete: kubeadmControlPlane %s rolloutAfter spec set %v", ownerRef.Name, *kcp.Spec.RolloutAfter))
+	} else {
+		r.Log.Info(fmt.Sprintf("failuredomain delete: kubeadmControlPlane %s rolloutAfter already set, skip", ownerRef.Name))
+	}
+	return ctrl.Result{}, nil
+}
+
+func checkClusterReady(r *CloudStackFailureDomainReconciliationRunner) (ctrl.Result, error) {
+	for _, condition := range r.CAPICluster.Status.Conditions {
+		if condition.Type == "ControlPlaneReady" && condition.Status == "False" {
```
maxdrib, line 330: let's use constants here as well

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,21 +140,215 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+			}
+			// etcdadmcluster rollout will trigger control plane rollout automatically.
+			// if etcd VM exists, trigger etcdadmcluster rollout is enough, no need to trigger control plane VM rollout.
+			if ref.Kind == "EtcdadmCluster" {
+				return ctrl.Result{}, nil
+			}
+		}
+	}
+	if !cpMachineFound {
+		return ctrl.Result{}, nil
```
maxdrib, line 302: do we need to log anything here?

controllers/cloudstackfailuredomain_controller.go
wanyufe: when etcdadmcluster's infrastructureTemplate cloudstack template get changed, a rollout will be triggered for etcdadmcluster. 

The initial settings are generated by eksa, but changing its settings later on seems fine - when eksa upgrade, it will generate a new cloudstack machine template.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -42,6 +42,13 @@ type CloudStackFailureDomainReconciler struct {
 	csCtrlrUtils.ReconcilerBase
 }
 
+const (
+	conditionTypeControlPlaneReady = "ControlPlaneReady"
```
maxdrib, line 46: nit: would we be able to use the constants from https://github.com/kubernetes-sigs/cluster-api/blob/6b5afa3d141da106ce03d163ebbe10a283b24d62/api/v1beta1/condition_consts.go#L78 and https://github.com/kubernetes-sigs/cluster-api/blob/6b5afa3d141da106ce03d163ebbe10a283b24d62/api/v1beta1/condition_consts.go#L24

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,188 @@ func setupMachineCRDs() {
...
+	// Fetch the CAPI Machine Deployment that was created.
+	key := client.ObjectKey{Namespace: dummies.ClusterNameSpace, Name: etcdadmCluster.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, etcdadmCluster)
+	}, timeout).Should(BeNil())
+}
+
+//
+//// setupFailuredomainCRDs creates a CloudstackFailuredomain.
+//func setupFailuredomainCRDs() {
```
maxdrib, line 489: should this commented code be removed?

controllers/controllers_suite_test.go
```diff
@@ -20,11 +20,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	etcdadmController "github.com/mrajashree/etcdadm-controller/api/v1beta1"
```
maxdrib, line 23: shouldn't this be under `aws` instead of `mrajashree`?

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +49,120 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
+			Eventually(func() bool {
+				key := client.ObjectKeyFromObject(dummies.CAPIMachineDeployment)
+				if err := k8sClient.Get(ctx, key, tempMd); err == nil {
+					_, found := tempMd.Spec.Template.Annotations["cluster.x-k8s.io/restartedAt"]
+					return found
+				}
+				return false
+			}, timeout).WithPolling(pollInterval).Should(BeTrue())
+		})
+		It("Should trigger kubeadmControlPlane rollout when control plane node exists failure domain.", func() {
```
maxdrib, line 82: I'm not sure I understand this test description. Could you rephrase it? Same with the next test

controllers/cloudstackfailuredomain_controller.go
wanyufe: No need to log anything since no CP VM in this failure domain is a normal situation. It is not an error or exceptional thing. 
I used to have something logged in this PR's draft version for easier search of logs.

rejoshed (commented):
For full understanding of the intent of this PR here is @maxdrib's excellent description.

> Initiate a rolling upgrade for the respective “manager” of the given machine type in order to wipe the machines from a given failuredomain. If a machine exists in an etcd cluster distributed across failure domains, for example, you need some way to move those machines to another failure domain before it can be deleted. The PR presents mechanisms for triggering this rolling upgrade for each type of machine - etcd, control plane, and worker node.


Personally, I don’t think it’s the right solution and feels like a hammer, but I have to acknowledge it does work and is a creative solution.

My critiques are:
- This triggers an unnecessary rolling update and if you’re only removing a single failure domain, now all of your machines across all of your cluster will get replaced.
- The code here is fairly complex. I think this could be rectified though.
- This more [tightly couples CAPC to ETCD ADM](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/156/files#diff-a1cc12a95fbe0a80302470b843b62b645345e5e781ab4e00a479f39aa7e47e12R183) which [when last discussed](https://amzn-aws.slack.com/archives/C02AG8JV5SP/p1661964911683889?thread_ts=1661867189.560149&cid=C02AG8JV5SP) we’d decided should be done via EKS-A patches.

The issue it’s solving AFAIK is that we still distribute worker nodes randomly and if all those worker nodes are on a single failure domain, then we don’t want to delete them all at once. Worst case is essentially that we delete those nodes all at once, and CAPC was running on those worker nodes while managing itself.

I’d propose two possible other solutions.
1). We only allow 1 deletion at a time — giving sufficient time for another machine to come up in that time.
2). We use verification webhooks to check that the entirety of a machine deployment hasn’t landed in a failure domain. (edited)


Also regarding ETCD machines,
> If a machine exists in an etcd cluster distributed across failure domains...
An external ETCD deployment should have multiple machines in an HA configuration already shouldn't it?

Plus, the latency burden encountered when placing multiple ETCD nodes in two different FailureDomains is questionable to begin with.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,25 +152,229 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+
+	result, err = triggerMachineDeploymentRollout(machines.Items, r)
+	if err != nil {
+		return result, err
+	}
+
+	if len(machines.Items) > 0 {
+		return r.RequeueWithMessage("failuredomain delete: FailureDomain still has machine(s) in it.", "failuredomain", r.ReconciliationSubject.Spec.Name)
+	}
+	return ctrl.Result{}, nil
```
rejoshed, line 180: If you make the trigger methods reconciliation runner methods than you can instead do.
``` Go
return r.RunStages(
   r.triggerA,
   r.triggerB,
   r.triggerC)
```

rejoshed: And actually anything of the form: `checkClusterReady(r)` can be done as `r.checkClusterReady` in a stage.

controllers/cloudstackmachine_controller.go
```diff
@@ -150,6 +150,29 @@ func (r *CloudStackMachineReconciliationRunner) SetFailureDomainOnCSMachine() (r
 	return ctrl.Result{}, nil
 }
 
+// SetFailureDomainOnCSMachine sets the failure domain the machine should launch in.
```
jweite-amazon, line 153: Func name in comment disagrees with actual func name.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,25 +152,229 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+					return ctrl.Result{}, err
+				}
+
+				newCsMachineTemplateName, err := cloneCsMachineTemplateWithNewName(r, machine.Namespace, csMachineTemplateName, ref.Name)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+
+				// patch etcdadmcluster to refer newly created cloudstack machine template, which will trigger an etcdadmcluster rolling upgrade.
+				if _, err := controllerutil.CreateOrPatch(r.RequestCtx, r.K8sClient, etcdadmCluster,
```
jweite-amazon, line 207: What deletes the original machine template?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,25 +152,229 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+	var cpMachine infrav1.CloudStackMachine
+	var ownerRef metav1.OwnerReference
+	cpMachineFound := false
+	for _, machine := range machines {
+		for _, ref := range machine.OwnerReferences {
+			if ref.Kind == kindKubeadmControlPlane {
+				cpMachineFound = true
+				cpMachine = machine
+				ownerRef = ref
+			}
```
jweite-amazon, line 276: It's surprising to me that this doesn't break on the first found CP machine.  Reading further, I see it wants to check whether there are any etcd machines, so it makes sense that it presses on.  I guess this means, then, that it acts on the *last* control plane machine found.  Which I imagine is fine.  But I'd suggest comments to make it clear that that's your intent, since that's a less typical approach.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,25 +152,229 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+				}
+
+				// add an annotation restartedAt in machine deployment if such one not already added
+				// this will trigger an immediate machine deployment rollout
+				patcher, err := patch.NewHelper(md, r.K8sClient)
+				if md.Spec.Template.Annotations == nil {
+					md.Spec.Template.Annotations = map[string]string{}
+				}
+				timeNowStr := time.Now().Add(time.Second * time.Duration(10)).Format(time.RFC3339)
+				if err != nil {
```
jweite-amazon, line 251: I think error check blocks should be immediately after the statement that sets the error variable.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -130,25 +152,229 @@ func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Resu
...
+					return ctrl.Result{}, err
+				}
+
+				// add an annotation restartedAt in machine deployment if such one not already added
+				// this will trigger an immediate machine deployment rollout
+				patcher, err := patch.NewHelper(md, r.K8sClient)
+				if md.Spec.Template.Annotations == nil {
+					md.Spec.Template.Annotations = map[string]string{}
+				}
+				timeNowStr := time.Now().Add(time.Second * time.Duration(10)).Format(time.RFC3339)
```
jweite-amazon, line 250: Is the 10s offset based on anything, and if so, are the any situations where this 10s offset might not be adequate?

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
 		return ph.Patch(ctx, dummies.CSMachine1, patch.WithStatusObservedGeneration{})
 	}, timeout).Should(Succeed())
 }
+
+// setupMachineCRDs creates a CAPI and CloudStack machine with an appropriate ownership ref between them.
```
jweite-amazon, line 437: Comment doesn't match function.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+			APIVersion: clusterv1.GroupVersion.String(),
+			Name:       CAPIMachine.Name,
+			UID:        "uniqueness",
+		})
+		return ph.Patch(ctx, CSMachine, patch.WithStatusObservedGeneration{})
+	}, timeout).Should(Succeed())
+}
+
+// setupCAPIMachineDeploymentCRDs creates a CAPI machine deployment.
+func setupCAPIMachineDeploymentCRDs(CAPIMachineDeployment *clusterv1.MachineDeployment) {
```
jweite-amazon, line 464: Since it only creates one CR, the function name should be singular.
```suggestion
func setupCAPIMachineDeploymentCRD(CAPIMachineDeployment *clusterv1.MachineDeployment) {
```

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+
+	// Fetch the CAPI Machine Deployment that was created.
+	key := client.ObjectKey{Namespace: dummies.ClusterNameSpace, Name: CAPIMachineDeployment.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CAPIMachineDeployment)
+	}, timeout).Should(BeNil())
+}
+
+// setupEtcdadmClusterCRDs create an etcdadm Cluster
+func setupEtcdadmClusterCRDs(etcdadmCluster *etcdadmController.EtcdadmCluster) {
```
jweite-amazon, line 476: Since it only creates one CR, the function name should be singular.
```suggestion
func setupEtcdadmClusterCRD(etcdadmCluster *etcdadmController.EtcdadmCluster) {
```

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+}
+
+// labelMachineFailuredomain add cloudstackfailuredomain info in the labels.
+func labelMachineFailuredomain(CSMachine *infrav1.CloudStackMachine, CSFailureDomain1 *infrav1.CloudStackFailureDomain) {
+	key := client.ObjectKey{Namespace: dummies.CSCluster.Namespace, Name: CSMachine.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachine)
+	}, timeout).Should(BeNil())
+
+	// Set owner ref from CAPI machine to CS machine and patch back the CS machine.
```
jweite-amazon, line 494: Comment doesn't seem accurate (wrt ownerRef setting).

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+		return k8sClient.Get(ctx, key, CAPIMachineDeployment)
+	}, timeout).Should(BeNil())
+}
+
+// setupEtcdadmClusterCRDs create an etcdadm Cluster
+func setupEtcdadmClusterCRDs(etcdadmCluster *etcdadmController.EtcdadmCluster) {
+	//  Create etcd adm cluster.
+	Ω(k8sClient.Create(ctx, etcdadmCluster)).Should(Succeed())
+
+	// Fetch the CAPI Machine Deployment that was created.
```
jweite-amazon, line 480: Comment inaccurate.  Fetching an etcdadmCluster.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+}
+
+// labelMachineDeploymentName add machine deployment info in the labels.
+func labelMachineDeploymentName(CSMachine *infrav1.CloudStackMachine, machineDeploymentName string) {
+	key := client.ObjectKey{Namespace: dummies.CSCluster.Namespace, Name: CSMachine.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachine)
+	}, timeout).Should(BeNil())
+
+	// Set owner ref from CAPI machine to CS machine and patch back the CS machine.
```
jweite-amazon, line 510: Comment inaccurate wrt ownerRef.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+	}, timeout).Should(Succeed())
+}
+
+func setMachineAnnotation(CSMachine *infrav1.CloudStackMachine, annotationKey string, annotationValue string) {
+	key := client.ObjectKey{Namespace: dummies.CSCluster.Namespace, Name: CSMachine.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachine)
+	}, timeout).Should(BeNil())
+
+	// Set owner ref from CAPI machine to CS machine and patch back the CS machine.
```
jweite-amazon, line 551: Comment inaccurate (wrt ownerRef).

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+	key := client.ObjectKey{Namespace: dummies.ClusterNameSpace, Name: CSMachineTemplate.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachineTemplate)
+	}, timeout).Should(BeNil())
+}
+
+func setMachineOwnerReference(CSMachine *infrav1.CloudStackMachine, ownerRef metav1.OwnerReference) {
+	key := client.ObjectKey{Namespace: dummies.CSCluster.Namespace, Name: CSMachine.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachine)
```
jweite-amazon, line 533: Updating the Machine object passed in by ref seems like an unexpected (to the caller) side-effect of calling this func.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+		Ω(err).ShouldNot(HaveOccurred())
+		CSMachine.OwnerReferences = append(CSMachine.OwnerReferences, ownerRef)
+		return ph.Patch(ctx, CSMachine, patch.WithStatusObservedGeneration{})
+	}, timeout).Should(Succeed())
+}
+
+func setMachineAnnotation(CSMachine *infrav1.CloudStackMachine, annotationKey string, annotationValue string) {
+	key := client.ObjectKey{Namespace: dummies.CSCluster.Namespace, Name: CSMachine.Name}
+	Eventually(func() error {
+		return k8sClient.Get(ctx, key, CSMachine)
```
jweite-amazon, line 548: Updating the Machine object passed in by ref seems like an unexpected (to the caller) side-effect of calling this func.

jweite-amazon:
Kudos for tackling a gnarly bit of work, with great logical clarity in the code @wanyufe.  

Like @rejoshed, I'm concerned the managing unstacked etcd machines is drawing CAPC into a responsibility it didn't establish.  I can see why it's practically necessary for the known use case, and it looks to me like the dependency is relatively soft (insofar as its mainly via kind string and unstructured).

controllers/cloudstackfailuredomain_controller.go
wanyufe: cloudstack machine templates keeps, nobody remove the old ones. I think this is true for cluster upgrade.

wanyufe: orignal machine template will not be deleted -- I think this is true for normal upgrade.

api/v1beta2/fakekindwithinfrastructuretemplate_types.go
```diff
@@ -0,0 +1,53 @@
...
+
+package v1beta2
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+//+kubebuilder:object:root=true
+//+kubebuilder:storageversion
+
+// FakeKindWithInfrastructure is the Schema for testing CRDs with infrastructureTemplate under spec
+type FakeKindWithInfrastructureTemplate struct {
```
maxdrib, line 25: why is this defined in the CAPC public API? I feel like it should be defined in a test package and not exposed to CAPC users.

maxdrib (commented):
Looks like some prow test is failing?

api/v1beta2/fakekindwithinfrastructuretemplate_types.go
wanyufe: I agree you on this. still trying to figure out way to put in a test package, it is harder than expected.
The challenge here: 

1. a CRD needs to be created so that our envtest can load it.
2. a scheme needs to addToScheme so that we have the map between GVK and go struct

Currently, our test package has not set it up that a deep_copy and scheme builder markers and so on, eventually, test will looks like a API. Still not certain if this is the correct path.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +433,137 @@ func setupMachineCRDs() {
...
+		CSMachine.Labels["cluster.x-k8s.io/deployment-name"] = machineDeploymentName
+		return ph.Patch(ctx, CSMachine, patch.WithStatusObservedGeneration{})
+	}, timeout).Should(Succeed())
+}
+
+func setCSMachineTemplateCRDs(CSMachineTemplate *infrav1.CloudStackMachineTemplate) {
+	//  Create cloudstack machine template.
+	Ω(k8sClient.Create(ctx, CSMachineTemplate)).Should(Succeed())
+
+	// Fetch the CAPI Machine Deployment that was created.
```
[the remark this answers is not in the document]
wanyufe: line 521 create, line 524 fetch it back to confirm it is created successfully.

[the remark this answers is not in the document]
wanyufe: corrected.

controllers/controllers_suite_test.go
wanyufe: corrected.

controllers/controllers_suite_test.go
wanyufe: corrected.

controllers/controllers_suite_test.go
wanyufe: corrected.

controllers/controllers_suite_test.go
wanyufe: corrected.

api/v1beta2/fakekindwithinfrastructuretemplate_types.go
wanyufe: I managed to move it to dummies package by solved above mentioned challenge.

rejoshed: Testing CAPC is fun huh?

controllers/cloudstackfailuredomain_controller.go
rejoshed: Curious... I too share your concern here @maxdrib.

How exactly is there no conflict? Does EKS-A somehow get feedback on this?

I would think modifying the (input) specs on most any of the CAPI/CAPC custom resources would be fought by EKS-A's incessant application of its spec. 

I like using incessant here. haha : p

test/e2e/config/cloudstack.yaml
```diff
@@ -134,6 +134,7 @@ variables:
   CLOUDSTACK_IMPOSSIBLE_CONTROL_PLANE_MACHINE_OFFERING: "Impossible Instance"
   CLOUDSTACK_WORKER_MACHINE_OFFERING: "Medium Instance"
   CLOUDSTACK_IMPOSSIBLE_WORKER_MACHINE_OFFERING: "Impossible Instance"
+  CLOUDSTACK_ETCD_MACHINE_OFFERING: "Large Instance"
```
maxdrib, line 137: why do we need this?

api/v1beta2/fakekindwithinfrastructuretemplate_types.go
wanyufe: I felt your pain.

test/e2e/config/cloudstack.yaml
wanyufe: it is not referred anymore. During previous commit, it is used to set up an etcd CSMachine offering.

controllers/controllers_suite_test.go
jweite-amazon: No, bad comment on my part.  Agreed that the function name should have made it clear that the CSMachine would be updated.

controllers/controllers_suite_test.go
jweite-amazon: Bad comment on my part.  Agreed that the function name should have made it clear that the CSMachine would be updated.

controllers/cloudstackfailuredomain_controller.go
rejoshed: Honestly, I think the problem this PR is aiming to solve needs to be discussed with all of the projects involved.

CAPI designers in particular may have a notion about failureDomain deletion, and if not, then we need to push them on why?

Either they fully support this feature or they don't. I don't like that we're in between two worlds.

We'll end up building something that half uses CAPI's failureDomain implementation and half doesn't.

Later CAPI may then fully commit or get out of this feature and where will that leave us?

rejoshed: But also CAPI, EKS-A, and ETCD owners need to probably find a way to trigger a rolling update in a generic way.

I think any controller that manages CAPI Machines should standardize on the way they manage those machines.

wanyufe: It kinds of like replicasets under deployment, each deployment changes will generate a new replicasets and it can be used for history record and roll back.

It does leak resource in etcd. The solution used in deployment is having a `revisionHistoryLimit` in deployment spec. 

So a similar solution for this is to have a `templateHistoryLimit` under somewhere (etcdadmcluster, capc's cloudstackmachinetemplate controller?)

SImply deleting the old cloudstackmachine template in above PR code is risky since cloudstackmachines are still referring them in `cloned-from`  labels.

rejoshed:
Went with another method for failure domain deletion.

PR #157, opened 2022-08-30
Revised doc files for V0.4.7

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Changes, primarily to getting-started.md, due to the change to yaml-secret-based credentialing.

*Testing performed:*


docs/book/src/development/common.md
```diff
@@ -31,7 +31,7 @@
     # Referring to a prerequisite capi-compatible image you've loaded into Apache CloudStack
     export CLOUDSTACK_TEMPLATE_NAME=kube-v1.20.10/ubuntu-2004
 
-    # The SSH KeyPair to log into the VM (Optional)
+    # The SSH KeyPair to log into the VM (Optional: use flavor *managed-ssh*)
```
maxdrib, line 34: speaking of flavors, I believe the template files currently are all based on the v1beta1 schema. Should we also bump them to the v1beta2 schema and explain how we're using failuredomains to support multi-endpoint somewhere?

docs/book/src/getting-started.md
```diff
@@ -9,32 +9,40 @@
...
         ```
         kind create cluster
         ```
+4. Set up Apache CloudStack credentials as a secret in the management cluster
+   - Create a file named `cloud-config.yaml` in the repo's root directory, substituting in your own environment's values
+       ```
+       apiVersion: v1
+       kind: Secret
+       metadata:
+         name: cloudStackCredentials
```
maxdrib, line 30: this name needs to be referenced in the corresponding failureDomain in the CloudStackCluster. We should mention that somewhere

docs/book/src/getting-started.md
```diff
@@ -9,32 +9,40 @@
...
+         namespace: default
+       type: Opaque
+       stringData:
+         api-key: <cloudstackApiKey>
+         secret-key: <cloudstackSecretKey>
+         api-url: <cloudstackApiUrl>
+         verify-ssl: "false"
```
maxdrib, line 37: I tried creating these texts with plaintext and I got this error 

```
Error from server (BadRequest): error when creating "capc_secret.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: decode base64: illegal base64 data at input byte 4, error found in #10 byte of ...|lient/api","secret-k|..., bigger context ...|VGc=","api-url":"http://10.11.0.2:8080/client/api","secret-key":"..."
```

I think they might have to be individual b64 encoded

docs/book/src/development/common.md
jweite-amazon: Template v1beta2 bump is in a separate [PR-158](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/158).

Agreed that failuredomains are worthy of documentation coverage.

docs/book/src/getting-started.md
jweite-amazon: Secret name *cloudStackCredentials* is a newly added default value for CLOUDSTACK_FD1_NAME in the templates.  No need for the end-user to coordinate these.

jweite-amazon: Changing to cloudstack-credentials because camel case isn't supported (per the below comment).

docs/book/src/getting-started.md
jweite-amazon: CamelCase not supported here.  Will change to cloudstack-credentials, here and in the templates.

docs/book/src/getting-started.md
jweite-amazon: The secret name used here is the newly added default secret name in the templates.  It should all connect without user involvement.

maxdrib (approved):

PR #141, opened 2022-07-29
e2e: Add kubernetes version upgrade test

davidjumani:
*Description of changes:*

Adds a Kubernetes version upgrade to the e2e test suite

*Testing performed:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #162, opened 2022-09-14
Revert "api: Allow changes to CloudstackMachineTemplate"

rejoshed:
Reverts kubernetes-sigs/cluster-api-provider-cloudstack#142

cloudstackmachinetemplates should be immutable per the CAPI spec.

It was mistakenly thought just modifying a template was sufficient to trigger a rolling update, but it's not and CAPI specifically documents that infra machine templates should be immutable.

See here for further details: https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/141#pullrequestreview-1104246170

### DIFF
--- a/api/v1beta2/cloudstackmachinetemplate_webhook.go
+++ b/api/v1beta2/cloudstackmachinetemplate_webhook.go
@@ -17,8 +17,11 @@ limitations under the License.
 package v1beta2
 
 import (
+	"fmt"
+	"reflect"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/webhookutil"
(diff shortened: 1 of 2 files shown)


api/v1beta2/cloudstackmachinetemplate_webhook.go
```diff
@@ -76,7 +79,30 @@ func (r *CloudStackMachineTemplate) ValidateCreate() error {
...
+
+	errorList := field.ErrorList(nil)
+	errorList = webhookutil.EnsureBothFieldsAreEqual(spec.Offering.ID, spec.Offering.Name, oldSpec.Offering.ID, oldSpec.Offering.Name, "offering", errorList)
+	errorList = webhookutil.EnsureBothFieldsAreEqual(spec.DiskOffering.ID, spec.DiskOffering.Name, oldSpec.DiskOffering.ID, oldSpec.DiskOffering.Name, "diskOffering", errorList)
+	errorList = webhookutil.EnsureStringFieldsAreEqual(spec.SSHKey, oldSpec.SSHKey, "sshkey", errorList)
+	errorList = webhookutil.EnsureBothFieldsAreEqual(spec.Template.ID, spec.Template.Name, oldSpec.Template.ID, oldSpec.Template.Name, "template", errorList)
+	errorList = webhookutil.EnsureStringStringMapFieldsAreEqual(&spec.Details, &oldSpec.Details, "details", errorList)
+	errorList = webhookutil.EnsureStringFieldsAreEqual(spec.Affinity, oldSpec.Affinity, "affinity", errorList)
+
+	if !reflect.DeepEqual(spec.AffinityGroupIDs, oldSpec.AffinityGroupIDs) { // Equivalent to other Ensure funcs.
```
maxdrib, line 101: best practice suggests avoiding use of reflect.DeepEqual. Could we define an Equal() function on this object and use that instead?

rejoshed: We could. But for now I simple did a reversion of the previous code.

If that's a blocker, then I'm fine with it changing to a defined equal, but I'll not be able to get to it asap.

I also don't necessarily agree with this. Defined equal functions sound nice, but in practice they can become outdated and in our case we truly want a template to have an immutable spec.

rejoshed: Further, we (and other providers) use DeepEqual in our watches too.

https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/a5fc9afbbbeaabcf0d6788c322f477da099da676/controllers/cloudstackcluster_controller.go#L177

And we don't seem to have any perf issues there either. Watches are definitely called more often than verification webhooks too.

I think DeepEqual has it's place.

maxdrib: The issue is type safety, especially after marshalling/unmarshalling. The full explanation is here: https://www.gojek.io/blog/relooking-at-golangs-reflect-deepequal. Maintaining Equal() functions is definitely a tradeoff, but those are readable and easy to interpret, whereas troubleshooting issues with reflect is not. Going forward, my vote would be in favor of equal operations to simplify debugging in case issues arise, but I'm not going to block this PR on it.

rejoshed: https://www.gojek.io/blog/relooking-at-golangs-reflect-deepequal

This is more to do with untyped JSON though. In this case and our other uses we already have a cloudstackmachinetemplate that's been de-serialized into a type.

rejoshed: I could see benefit to a custom equals in that instead of saying 

> You can't change cloudstackmachinetemplates"

we'd say

> You cannot change the _offering_ of a cloudstackmachinetemplate. 

As you mentioned the error messages would be more detailed. That said, I haven't dug in yet, but I'm guessing that there is a way to ask DeepEqual which fields were not equal and plop those into an error message.

PR #164, opened 2022-09-23
Failuredomain deletion

wanyufe:
*Issue #, if available:*
Update failure domain remving logic about how to delete CAPI machines.
*Description of changes:*
Previous implementation simply delete CAPI machines in deleted failure domain who can potential caused issues when machinesets, kubeadmcontrolplane and etcdadmcluster losing majority or capacity.
This commit uses the logic to only delete one CAPI machine a time and only delete machines when replicas are more than 1.
*Testing performed:*
Manual testing and unit testing


controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
 	)
 }
 
-// ClearMachines checks for any machines in failure domain, deletes the CAPI machine for any still in FailureDomain,
-// and requeus until all CloudStack machines are cleared from the FailureDomain.
-func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Result, error) {
+// GetAllMachinesInFailureDomain get all cloudstack machines deployed in this failure domain.
+// machines are sorted by name so that it can be processed one by one in a determined order.
```
rejoshed, line 138:
```suggestion
// GetAllMachinesInFailureDomain returns all cloudstackmachines deployed in this failure domain sorted by name.
```

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+	r.Machines = items
+	return ctrl.Result{}, nil
+}
+
+// AllMachinesCanBeCleared checks for each machine in failure domain, check if it is possible to delete it.
+// if machine is the only machine in worker node group, it cannot be deleted.
+// if machine is the only machine in control plane, it cannot be deleted.
+// if machine is the only machine in etcdadmcluster, it cannot be deleted.
+// if deletes the CAPI machine for any still in FailureDomain,
+// and requeus until all CloudStack machines are cleared from the FailureDomain.
```
rejoshed, line 157: The comments and naming here is a bit weird.

All machines can be cleared seems like it should return a boolean value and have no side effects.

Yet according to the comment, this method deletes a machine if necessary and requeues.

I would change the name to ClearMachines or something if it is doing so. Then I would have another method that to call after called something like ReueueIfMachinesPresent.

That way it's clear what each method is doing.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+	// check CAPI machines for CloudStack machines found.
+	for _, machine := range r.Machines {
 		for _, ref := range machine.OwnerReferences {
+			if ref.Kind != "Machine" {
+				owner := &unstructured.Unstructured{}
+				owner.SetGroupVersionKind(schema.FromAPIVersionAndKind(ref.APIVersion, ref.Kind))
+				if err := r.K8sClient.Get(r.RequestCtx, client.ObjectKey{Namespace: machine.Namespace, Name: ref.Name}, owner); err != nil {
+					return ctrl.Result{}, err
+				}
+				specReplicas, statusReplicas, err := replicasLargerThanOne(owner, ref.Name, machine.Name)
```
rejoshed, line 168: The method naming here doesn't make too much sense to me.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+				if found && statusReadyReplicas != statusReplicas {
+					return r.RequeueWithMessage("status.replicas <> status.readyReplicas, ", "owner", ref.Name, "status.replicas", statusReplicas, "status.readyReplicas", statusReadyReplicas)
+				}
+			}
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func replicasLargerThanOne(owner *unstructured.Unstructured, ownerName, machineName string) (int64, int64, error) {
```
rejoshed, line 197: I think this method should be broken into two. The first should be fetching the attributes/fields, and the other two should be checking the numbers themselves.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+				specReplicas, statusReplicas, err := replicasLargerThanOne(owner, ref.Name, machine.Name)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				if specReplicas != statusReplicas {
+					return r.RequeueWithMessage("spec.replicas <> status.replicas, ", "owner", ref.Name, "spec.replicas", specReplicas, "status.replicas", statusReplicas)
+				}
+
+				statusReady, found, err := unstructured.NestedBool(owner.Object, "status", "ready")
+				if found && err != nil {
```
rejoshed, line 177: I'm confused. Why are we checking found here?

Shouldn't we return if an error was encountered regardless?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+					return ctrl.Result{}, err
+				}
+				if found && !statusReady {
```
rejoshed, line 180: Can you explain why we're concerned with the owner's readiness?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+					return r.RequeueWithMessage("status.ready not true, ", "owner", ref.Name)
+				}
+
+				statusReadyReplicas, found, err := unstructured.NestedInt64(owner.Object, "status", "readyReplicas")
```
rejoshed, line 184: Since we need all these fields, why not put all of the logic to fetch them into one method and handle the logic on these fields into another.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+				if found && err != nil {
+					return ctrl.Result{}, err
+				}
+				if found && statusReadyReplicas != statusReplicas {
+					return r.RequeueWithMessage("status.replicas <> status.readyReplicas, ", "owner", ref.Name, "status.replicas", statusReplicas, "status.readyReplicas", statusReadyReplicas)
+				}
+			}
+		}
+	}
+	return ctrl.Result{}, nil
```
rejoshed, line 194: Here we seem to exit the method, but the method hasn't called to delete a CAPI machine AFAICT. Yet the comment indicates that it should?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -123,29 +134,109 @@ func (r *CloudStackFailureDomainReconciliationRunner) ReconcileDelete() (ctrl.Re
...
+	if specReplicas < 2 {
+		return specReplicas, statusReplicas, errors.Errorf("spec.replicas < 2 in %s, %s cannot be moved away from failure domain", ownerName, machineName)
+	}
+
+	return specReplicas, statusReplicas, nil
+}
+
+// ClearMachines deletes the CAPI machine in FailureDomain.
+func (r *CloudStackFailureDomainReconciliationRunner) ClearMachines() (ctrl.Result, error) {
+	for _, csMachine := range r.Machines {
```
rejoshed, line 223: It doesn't really matter, but since we're only deleting one machine at a time, instead of a for loop we could do.

``` go
if len(r.Machines) > 0 {
    ...
    do single machine deletion process
    ...
}
```

Which might be slightly clearer than a for loop, but does exactly the same thing.

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +48,200 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
+			Eventually(func() bool {
+				key := client.ObjectKey{Namespace: dummies.ClusterNameSpace, Name: dummies.CAPIMachine.Name}
+				if err := k8sClient.Get(ctx, key, CAPIMachine); err == nil {
+					return CAPIMachine.DeletionTimestamp.IsZero()
+				}
+				return false
+			}, timeout).WithPolling(pollInterval).Should(BeTrue())
+		})
+
+		It("Should return error if status.ready is false.", func() {
```
rejoshed, line 109: With the number of spec inputs here, I'm pretty sure these specs could be condensed into a table.

https://onsi.github.io/ginkgo/#table-specs

Would probably make the whole set of specs much clearer and greatly reduce code repetition.

test/dummies/v1beta2/vars.go
```diff
@@ -130,6 +137,95 @@ func CAPCNetToCSAPINet(net *infrav1.Network) *csapi.Network {
 	}
 }
 
+func setCSMachineOwnerCrd() {
+	CSMachineOwnerCRD = &apiextensionsv1.CustomResourceDefinition{
```
rejoshed, line 141: I feel like this could have just been imported instead of redefined?

rejoshed: As in this could just be a machineset or similar imported from CAPI instead of creating your own.

test/fakes/CloudStackMachineOwner_types.go
```diff
@@ -0,0 +1,57 @@
...
+
+package fakes
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+//+kubebuilder:object:root=true
+//+kubebuilder:storageversion
+//+kubebuilder:subresource:status
```
rejoshed, line 22: Yeah, I think I would just import a CAPI machineset instead of rewriting an analog of it here. This adds a lot of code to think about, but functions identically as far as I can tell.

rejoshed (changes requested):
Please review my comments. Thanks for putting this together!

controllers/cloudstackfailuredomain_controller.go
wanyufe: the method name can be better. It can be renamed to `requeueIfMachineCannotBeRemoved`. Comments are also updated to reflect meaning of method.

controllers/cloudstackfailuredomain_controller.go
wanyufe: breaking this method into two or extract the logic out will trigger go warning about parent method complexity -- too much if in one method.

Renamed method to `getSpecAndStatusReplicas`

controllers/cloudstackfailuredomain_controller.go
wanyufe: I removed those comments. The comments are for background knowledge instead of what the method is doing.

test/dummies/v1beta2/vars.go
wanyufe: As I mentioned in above response, even machineset, kubeadmControlPlane and etcdadmCluster have similar pattern, like spec.replicas, status.replicas, they also have difference. For example, machineset does not have status.ready and etcdadmcluster does not have status.readyReplicas. kubeadmControlPlane has all fields but they are required fields and could not test `not found` logic.

controllers/cloudstackfailuredomain_controller.go
rejoshed: AFAICT this isn't about being `ready` though. If there was an error and it was found then we return the error, but if there was an error and it wasn't found we continue?

controllers/cloudstackfailuredomain_controller.go
rejoshed: Sure, but we could put all the field fetching logic into one method with sub-methods, and then all the logic that operates on those fields into another. This is a bit too hard to follow.

controllers/cloudstackfailuredomain_controller.go
rejoshed: Maybe we should call it `RequeueIfAnyMachineCannotBeRemoved`?

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +48,83 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
-					return tempfd.Status.Ready
-				}
-				return false
-			}, timeout).WithPolling(pollInterval).Should(BeTrue())
 		})
+
+		It("Should set failure domain Status.Ready to true.", func() {
+			assertFailureDomainCreated()
+		})
+		It("Should delete failure domain if no VM under this failure domain.", func() {
```
rejoshed, line 56: Let's be more clear about the description of the test here. Technically the failureDomain has a deletion timestamp, and the real crutch of the test is to check that the finalizer is removed.

Maybe something like `It("Should remove failure domain finalizer when no VMs are placed in it.", ...`

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +48,83 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
-				key := client.ObjectKeyFromObject(dummies.CSFailureDomain1)
-				if err := k8sClient.Get(ctx, key, tempfd); err == nil {
-					return tempfd.Status.Ready
-				}
-				return false
-			}, timeout).WithPolling(pollInterval).Should(BeTrue())
 		})
+
+		It("Should set failure domain Status.Ready to true.", func() {
+			assertFailureDomainCreated()
```
rejoshed, line 54: `assertFailureDomainCreated` appears in the beginning of every one of these tests. Maybe we should just move it to the before function?

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +48,83 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
+	Eventually(func() bool {
+		key := client.ObjectKeyFromObject(dummies.CSFailureDomain1)
+		if err := k8sClient.Get(ctx, key, tempfd); err != nil {
+			return true
+		}
+		return false
+	}, timeout).WithPolling(pollInterval).Should(BeTrue())
+}
+
+func boolPointer(b bool) *bool {
```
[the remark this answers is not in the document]
rejoshed: Plus I think you could just take the address of the args passed in when you call the `setCSMachineOwnerCRD` and other methods so you could just skip all this pointy-ness anyway.

controllers/cloudstackfailuredomain_controller_test.go
```diff
@@ -49,14 +48,83 @@ var _ = Describe("CloudStackFailureDomainReconciler", func() {
...
+			// should not delete if condition not met
+			Entry("Should return error if status.replicas < spec.replicas", false, int32Pointer(2), int32Pointer(1), int32Pointer(1), boolPointer(true)),
+			Entry("Should return error if spec.replicas < 2", false, int32Pointer(1), int32Pointer(1), int32Pointer(1), boolPointer(true)),
+			Entry("Should return error if status.ready is false", false, int32Pointer(2), int32Pointer(2), int32Pointer(2), boolPointer(false)),
+			Entry("Should return error if status.readyReplicas <> status.replicas", false, int32Pointer(2), int32Pointer(2), int32Pointer(1), boolPointer(true)),
+		)
 	})
 })
+
+func assertFailureDomainCreated() {
```
rejoshed, line 102: I'm fine with this as is, but my preference would be to wrap the getter and checker in an Eventually outside of a function like this. That way the eventually is more obvious when reading through a spec.

This kind of hides the eventually in a method and makes it slightly more opaque that the code may just hang at this spot until an eventual failure occurs.

controllers/controllers_suite_test.go
```diff
@@ -142,6 +143,7 @@ var _ = BeforeSuite(func() {
 
 	Ω(infrav1.AddToScheme(scheme.Scheme)).Should(Succeed())
 	Ω(clusterv1.AddToScheme(scheme.Scheme)).Should(Succeed())
+	Ω(fakes.AddToScheme(scheme.Scheme)).Should(Succeed())
```
rejoshed, line 146: So, I still don't understand the need for a `fakes` schema. I know you put a lot of work into this, but I think this was a bit of a miscommunication on my part. I think that just importing a machineset or equivalent would be completely sufficient here.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +373,82 @@ func setupMachineCRDs() {
...
+			Kind:       "Machine",
+			APIVersion: clusterv1.GroupVersion.String(),
+			Name:       CAPIMachine.Name,
+			UID:        "uniqueness",
+		})
+		return ph.Patch(ctx, CSMachine, patch.WithStatusObservedGeneration{})
+	}, timeout).Should(Succeed())
+}
+
+func setMachineOwnerReference(CSMachine *infrav1.CloudStackMachine, ownerRef metav1.OwnerReference) {
```
rejoshed, line 425: Could these methods be abstracted to operate on `client.Objects` instead of the specific types? We kind of do that already in a few places in our reconciler base methods.

controllers/controllers_suite_test.go
```diff
@@ -371,3 +373,82 @@ func setupMachineCRDs() {
...
+	Eventually(func() error {
+		owner.Status.Ready = statusReady
+		owner.Status.Replicas = statusReplicas
+		owner.Status.ReadyReplicas = statusReadyReplicas
+		return k8sClient.Status().Update(ctx, owner)
+	}, timeout).Should(BeNil())
+}
+
+// setCAPIMachineAndCSMachineCRDs creates a CAPI and CloudStack machine with an appropriate ownership ref between them.
+func setCAPIMachineAndCSMachineCRDs(CSMachine *infrav1.CloudStackMachine, CAPIMachine *clusterv1.Machine) {
```
rejoshed, line 394: I sort of think these methods should be broken into another file. I'd like one file that just sets up the base objects an another to operate on their relationships. I'm fine with it as is if it's too much trouble, but I think that would be nice organizationally.

test/fakes/CloudStackMachineOwner_types.go
rejoshed: But I think we could just import objects that either do or don't have those field?

controllers/controllers_suite_test.go
wanyufe: for example, we use machineset, but it does not status.ready field. And we could not the logic when ready is false/true case.

controllers/cloudstackfailuredomain_controller.go
jweite-amazon: +1

PR #168, opened 2022-09-30
Remove checking name when looking for a certain error message because…

wongni:
… namespace is unique enough in e2e testing

*Issue #, if available:*
Three e2e test cases are failing
```
Should fail due to the specified domain is not found [TC4b] [It]
Should fail due to the specified account is not found [TC4a] [It]
Should fail due to the specified zone is not found [TC3] [It]
```

Looks like the name in the error log printed from failure domain controller is not something the test expect (cluster name). 
```
E0930 15:34:33.597419       1 controller.go:317] controller/cloudstackfailuredomain "msg"="Reconciler error" "error"="resolving account accountXXXX details: could not find account accountXXXX" "name"="470ad1cb9d240bf93ab6fc6586b0052c" "namespace"="invalid-resource-mqt4sb" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="CloudStackFailureDomain" 
```

*Description of changes:*
This PR removes checking the name field when looking for a certain error message because the namespace only is unique enough in e2e testing.

*Testing performed:*

### DIFF
--- a/test/e2e/invalid_resource.go
+++ b/test/e2e/invalid_resource.go
@@ -227,8 +227,7 @@ func errorExistsInLog(logFolder string, expectedError string) (bool, error) {
 			logLines := strings.Split(string(log), "\n")
 			for _, line := range logLines {
 				if strings.Contains(line, expectedError) &&
-					strings.Contains(line, clusterResources.Cluster.Namespace) &&
-					strings.Contains(line, clusterResources.Cluster.Name) {
+					strings.Contains(line, clusterResources.Cluster.Namespace) {
 					Byf("Found %q error", expectedError)
 					return expectedErrorFound
 				}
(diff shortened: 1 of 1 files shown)


test/e2e/invalid_resource.go
```diff
@@ -227,8 +227,7 @@ func errorExistsInLog(logFolder string, expectedError string) (bool, error) {
 			logLines := strings.Split(string(log), "\n")
 			for _, line := range logLines {
 				if strings.Contains(line, expectedError) &&
-					strings.Contains(line, clusterResources.Cluster.Namespace) &&
-					strings.Contains(line, clusterResources.Cluster.Name) {
+					strings.Contains(line, clusterResources.Cluster.Namespace) {
```
jweite-amazon, line 230: Please edit the comment for this function, as it still states that it its confirming cluster name (line 215).  Consider adding to the comment to indicate why we can't check for name.

PR #171, opened 2022-10-11
Set owner reference to endpoint secret with the cluster if the secret…

wongni:
*Issue #, if available:* #172

*Description of changes:*
When a secret is dedicated to a cluster by labeling the secret with the cluster name and it's not already owned by a cluster CAPC sets the secret's owner reference with the target cluster. 

*Testing performed:*
TBD

### DIFF
--- a/controllers/cloudstackfailuredomain_controller.go
+++ b/controllers/cloudstackfailuredomain_controller.go
@@ -18,8 +18,11 @@ package controllers
 
 import (
 	"context"
+	"strings"
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -76,6 +79,10 @@ func (r *CloudStackFailureDomainReconciliationRunner) Reconcile() (retRes ctrl.R
 	// Prevent premature deletion.
 	controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.FailureDomainFinalizer)
 
+	if err := r.SetOwnerReferenceToEndpointSecret(); err != nil {
+		return r.RequeueWithMessage("Couldn't set owner reference for endpoint secret")
+	}
+
 	// Start by purely data fetching information about the zone and specified network.
 	if err := r.CSUser.ResolveZone(&r.ReconciliationSubject.Spec.Zone); err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "resolving CloudStack zone information")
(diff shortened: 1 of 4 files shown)


controllers/cloudstackfailuredomain_controller.go
```diff
@@ -107,6 +114,43 @@ func (r *CloudStackFailureDomainReconciliationRunner) Reconcile() (retRes ctrl.R
...
+	if err != nil {
+		return err
+	}
+	if MaySetOwnerReference(secret, r.CAPICluster) {
+		return r.Patcher.Patch(r.RequestCtx, secret)
+	}
+	return nil
+}
+
+func MaySetOwnerReference(secret *corev1.Secret, cluster *clusterv1.Cluster) bool {
```
jweite-amazon, line 128: The name of this function makes me think it's const (i.e., returning a bool indicating whether you may set the owner reference).  Since it's not, I'd prefer a name that indicates so, maybe OptionallySetOwnerReference.

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -76,6 +79,10 @@ func (r *CloudStackFailureDomainReconciliationRunner) Reconcile() (retRes ctrl.R
 	// Prevent premature deletion.
 	controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.FailureDomainFinalizer)
 
+	if err := r.SetOwnerReferenceToEndpointSecret(); err != nil {
+		return r.RequeueWithMessage("Couldn't set owner reference for endpoint secret")
```
maxdrib, line 83: Could you include the error message in the requeue message?

controllers/cloudstackfailuredomain_controller.go
```diff
@@ -107,6 +114,43 @@ func (r *CloudStackFailureDomainReconciliationRunner) Reconcile() (retRes ctrl.R
...
+			break
+		}
+	}
+	if needSetOwner {
+		secret.OwnerReferences = append(secret.OwnerReferences,
+			metav1.OwnerReference{
+				Name:       cluster.Name,
+				Kind:       cluster.Kind,
+				APIVersion: cluster.APIVersion,
+				UID:        cluster.UID,
```
maxdrib, line 148: I'm not a fan of modifying input parameters. Can we just return `needSetOwner` and modify the secret in the client?

maxdrib (commented):
A couple high level notes
1. How does this work if clusters are sharing secrets? 
2. I suspect there is some documentation which needs to be updated. Would you create an issue to do that?

wongni:
> A couple high level notes
> 
>     1. How does this work if clusters are sharing secrets?

If a secret is shared by multiple clusters it will not have the cluster name label. Therefore, any owner references wouldn't be set. For a certain customer, each workload cluster has a dedicated secret with the cluster name label. In this case, this PR sets the secret's owner to the corresponding cluster. 

> 
>     2. I suspect there is some documentation which needs to be updated. Would you create an issue to do that?

Sure.

wongni:
Closing because there is a better way: https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/172#issuecomment-1274998049

PR #173, opened 2022-10-11
Added least privilege role description to docs (book).

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Book changes: new page describing least-privilege ACS role required to successfully run CAPC

*Testing performed:*
Ran e2e tests against ACS DomainAdmin type accounts having a role with the documented privileges (rules).

### DIFF
--- a/docs/book/src/topics/cloudstack-permissions.md
+++ b/docs/book/src/topics/cloudstack-permissions.md
@@ -1,0 +1,42 @@
+# CloudStack Permissions for CAPC
+
+The account that CAPC runs under must minimally be a Domain Admin type account with a role offering the following permissions
+
+* assignToLoadBalancerRule
+* associateIpAddress
+* createAffinityGroup
+* createEgressFirewallRule
+* createLoadBalancerRule
+* createNetwork
+* createTags
+* deleteAffinityGroup
+* deleteNetwork
+* deleteTags
+* deployVirtualMachine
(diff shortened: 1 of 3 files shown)


wongni (approved):

docs/book/src/topics/cloudstack-permissions.md
```diff
@@ -0,0 +1,42 @@
...
+* listVirtualMachines
+* listVirtualMachinesMetrics
+* listVolumes
+* listZones
+* queryAsyncJobResult
+* startVirtualMachine
+* stopVirtualMachine
+* updateVMAffinityGroup
+
+This permission set has been verified to successfully run the CAPC E2E test suite.
```
maxdrib, line 42: I suspect this will get stale relatively quickly. Maybe add a date for when this was last assessed? Ideally we'd have a bootstrap script to create the user running the e2e tests and grant them all these permissions, and we'd just be able to link to that script, but I'm guessing that's too much

jweite-amazon: Agreed.  I've added the date to the doc.  The script would be great, but I'm not certain I can do that now.  Best bet, though, is to develop under an ACS account/user with these permissions, so one would immediately know if one was stepping beyond the established least priv.

maxdrib (approved):
Lgtm

PR #177, opened 2022-10-13
Fixing Makefile bugs

maxdrib:
*Issue #, if available:*

*Description of changes:*
When attempting to build the latest release inside https://github.com/aws/eks-anywhere-build-tooling/pull/1427, there were several failures related to path issues and shell flags. The eks-anywhere repo checks out the CAPC repo inside it, and then runs the build from the top level so using relative paths won't work. Additionally, the Makefie shell commands were showing

```
bash: - : invalid option
```

with the shell flags enabled so I'm attempting to disable them.

*Testing performed:*
Initial failure observed in [prow](https://prow.eks.amazonaws.com/view/s3/prowpresubmitsdataclusterstack-prowbucket7c73355c-vfwwxd2eb4gp/pr-logs/pull/aws_eks-anywhere-build-tooling/1427/cluster-api-provider-cloudstack-tooling-presubmit/1580593304205004800). I was able to create a docker container with the same image running locally and test the same build command and I got to the point of pulling from the go proxy.

On my mac, I also tested the docs targets and was able to `build` and `serve` the docs locally. CAPC unit tests pass. CAPC builds. Docker image builds successfully with `docker-build` and `docker-push` targets.


wongni (approved):

common.mk
```diff
@@ -12,16 +12,16 @@
...
 # limitations under the License.
 
-include $(ROOT_DIR_RELATIVE)/versions.mk
+include $(REPO_ROOT)/versions.mk
 
 # Ensure Make is run with bash shell as some syntax below is bash-specific
 SHELL:=bash
 .ONESHELL:
-.SHELLFLAGS := -eu -o pipefail -c
+# .SHELLFLAGS := -eu -o pipefail -c
```
maxdrib, line 20: Somehow, these SHELLFLAGS being set was leading to failures every time this and other Makefiles would call the `shell` command. I do not fully understand why, but without them it works.

Makefile
```diff
@@ -12,18 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ROOT_DIR_RELATIVE := .
+export REPO_ROOT := $(shell git rev-parse --show-toplevel)
```
rejoshed, line 15: David made an argument about this being necessary as sometimes CAPC and other providers are included in other projects (Potentially CAPI itself). Using `ROOT_DIR_RELATIVE` is supposed to allow for this. I'm not sure we should remove this without fully understanding the implications.

maxdrib: I think @davidjumani may be in a conference and unavailable. 

The CAPC repo is also checked out as a subdirectory in the EKS-A build-tooling repo: https://github.com/aws/eks-anywhere-build-tooling/tree/main/projects/kubernetes-sigs/cluster-api-provider-cloudstack. When we build it, we clone the CAPC repo and build it below with make -C. I feel like using git rev-parse is safer than hardcoding `.`, which becomes ambiguous if you're running the build from a different directory

common.mk
rejoshed: [explanation of shell flags](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425)

-e exit if a failure occurs instead of chugging along
-u exit if an undefined variable is expanded/used
-pipefail exit if one command piped into another fails, so `false | true` should exit. This doesn't mean that `false || true` does though.

-c typically is used with a string thereafter so the shell executes the string as a command once it starts.

I'm not sure it's needed here, or when it was added, but maybe it is.

Did you try removing one flag at a time to see which flag in particular was causing the issue?

I don't think we should just blanket comment them all out.

Makefile
rejoshed: I think what David meant was that CAPC was used as a [submodule](https://www.git-scm.com/book/en/v2/Git-Tools-Submodules)

In that case the repo root according to git would no longer be the CAPC repo's root.

I'm not that familiar with submodules or David's reasoning/use case. So, if you think we really should remove this, then I suppose I'm fine with it. We've certainly released just fine without this change before. Of course most of our previous releases weren't actually functional from a `clusterctl init` standpoint.

common.mk
davidjumani: I agree. I've also spoken with @jweite-amazon and noted that these failures are OS specific

Makefile
davidjumani: `REPO_ROOT` is already defined in https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/main/Makefile#L26

`ROOT_DIR_RELATIVE` uses a similar pattern defined in other providers as well
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/Makefile#L15
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/Makefile#L15

I think this was needed for building the books, or something similar. Will get back to you on that.

common.mk
maxdrib: Thanks everybody, moving to using `-euc` instead to avoid spaces, which are not handled properly in the linux bash.

Makefile
maxdrib: >REPO_ROOT is already defined in https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/main/Makefile#L26, maybe it can be reused ?

This is where I got the repo root from. It's deleted in line 26 below

>Another reason was to align it with other providers. Eg:
ROOT_DIR_RELATIVE uses a similar pattern defined in other providers as well

I'm looking at CAPV and I see they have a fancy way to identify the root as well (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/Makefile#L41) so I think using git repo root should be appropriate here, or maybe vsphere's way.

wongni (approved):

PR #180, opened 2022-10-15
Update failuredomain regexp and prevent race condition while removing failure domain

wanyufe:
*Issue #, if available:*

Add cluster condition status checking in failure domain removing logic.
Prevent one race condition that machine can be allocated to a failure domain that just being deleted. 
*Description of changes:*
Enhance failure domain removing logic.
Don't remove machine when cluster status is ready.
Don't deploy VM in a failuredomain that is deleted.
*Testing performed:*
Unit testing and manual testing.

### DIFF
--- a/controllers/cloudstackfailuredomain_controller.go
+++ b/controllers/cloudstackfailuredomain_controller.go
@@ -157,6 +157,9 @@ func (r *CloudStackFailureDomainReconciliationRunner) GetAllMachinesInFailureDom
 // RequeueIfClusterNotReady check cluster to see if there is any rolling update going on.
 func (r *CloudStackFailureDomainReconciliationRunner) RequeueIfClusterNotReady() (ctrl.Result, error) {
 	if len(r.Machines) > 0 {
+		if !r.CAPICluster.DeletionTimestamp.IsZero() {
+			return ctrl.Result{}, nil
+		}
 		for _, condition := range r.CAPICluster.Status.Conditions {
 			if condition.Type == conditionTypeReady && condition.Status == conditionStatusFalse {
 				return r.RequeueWithMessage("cluster status not ready,")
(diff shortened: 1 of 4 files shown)


controllers/cloudstackfailuredomain_controller.go
```diff
@@ -148,6 +154,21 @@ func (r *CloudStackFailureDomainReconciliationRunner) GetAllMachinesInFailureDom
 	return ctrl.Result{}, nil
 }
 
+// RequeueIfClusterNotReady check cluster to see if there is any rolling update going on.
+func (r *CloudStackFailureDomainReconciliationRunner) RequeueIfClusterNotReady() (ctrl.Result, error) {
```
rejoshed, line 158: Interesting that there are two ways to check if a cluster is ready. Both the `status.Ready` and the `status.ControlPlaneReadyCondition`?

Are the two set under different circumstances?

Are we sure we want to prevent deletion if the control plane is in a not ready state? For example, what if it's not ready because it's deleting?

I'm sure that's not an issue as you must have ran this and it was fine.


In general though, could you please add some background (in comments is fine) on why this is necessary and what the difference between `status.Ready==true` and  `conditions.ControlPlaneReadyConditions` are?

wanyufe:
```Are we sure we want to prevent deletion if the control plane is in a not ready state? For example, what if it's not ready because it's deleting?
```
If that happens, that will be temporary since once deleting is done, new VM will be created due to reconcile and control plane will be ready again.

wanyufe: If cluster is being deleted, a boot strap (kind) cluster will be used to delete current cluster and cluster will be paused while being deleted.

rejoshed: Thank you for adding that conditional. Will merge now.

rejoshed (approved):

PR #185, opened 2022-10-26
Fix Makefile to run e2e

wongni:
*Issue #, if available:*

*Description of changes:*
Running e2e test fails with the following error. 
```
      failed to execute kustomize: : exec: "kustomize": executable file not found in $PATH
```

*Testing performed:*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ run: generate-deepcopy generate-conversion ## Run a controller from your host.
 
 .PHONY: deploy
 deploy: generate-deepcopy generate-manifests $(KUSTOMIZE) ## Deploy controller to the K8s cluster specified in ~/.kube/config.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,7 +75,7 @@ endif
 export ACK_GINKGO_DEPRECATIONS := 1.16.5
 export ACK_GINKGO_RC=true
 
-export PATH := $(REPO_ROOT)/$(TOOLS_BIN_DIR):$(PATH)
+export PATH := $(TOOLS_BIN_DIR):$(PATH)
```
maxdrib, line 78: I think there may be a similar change required on line 202. Would you mind looking at that, since we're fixing up the Makefile?

wongni: Good point. Fixed.

maxdrib (approved):
/lgtm

PR #187, opened 2022-11-01
Test CI integration

davidjumani:
*Issue #, if available:*

*Description of changes:*

*Testing performed:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,3 +199,4 @@ All the CAPC contributors:
 [k1.23-u20-vhd-md5]: http://packages.shapeblue.com/cluster-api-provider-cloudstack/images/xen/ubuntu-2004-kube-v1.23.3-xen.vhd.bz2.md5
 [k1.24-u20-vhd]:     http://packages.shapeblue.com/cluster-api-provider-cloudstack/images/xen/ubuntu-2004-kube-v1.24.7-xen.vhd.bz2
 [k1.24-u20-vhd-md5]: http://packages.shapeblue.com/cluster-api-provider-cloudstack/images/xen/ubuntu-2004-kube-v1.24.7-xen.vhd.bz2.md5
+


rohityadavcloud:
ping @davidjumani can you advise if this can be closed (the integration is working?)

rohityadavcloud:
Instead of the /run.... syntax, we're exploring once we fix the E2E tests to use our `@blueorangutan` bot to kick tests directly using a comment-based approach (which is what we use in the cloudstack repos), ShapeBlue hosted lab env allowing us to test multiple versions of CloudStack against different hypervisors.
/cc @weizhouapache @vishesh92 @davidjumani @g-gaston @chrisdoherty4 @hrak 

I think it would help us validate PRs faster once the E2E tests are fixed and the CI integration is resumed. Let me know if there are any concerns.

g-gaston:
> Instead of the /run.... syntax, we're exploring once we fix the E2E tests to use our `@blueorangutan` bot to kick tests directly using a comment-based approach (which is what we use in the cloudstack repos), ShapeBlue hosted lab env allowing us to test multiple versions of CloudStack against different hypervisors. /cc @weizhouapache @vishesh92 @davidjumani @g-gaston @chrisdoherty4 @hrak
> 
> I think it would help us validate PRs faster once the E2E tests are fixed and the CI integration is resumed. Let me know if there are any concerns.

@rohityadavcloud

In the short term, I'm down for whatever is easier and give us e2e tests so we can release with confidence.

The only problem I see is how do we integrate that with current k8s org permissions.
We shouldn't allow non members to trigger the tests (for security reasons). Prow takes care of this for us with `ok-to-test`, etc. How would that work with the bot your are proposing?

Also, what kind of visibility over the errors does his bot have? I wonder what kind of debugging experience contributors (without direct access to the test env) will have.

PR #191, opened 2022-11-11
fix bug of orphaned CS machine

wanyufe:
*Issue #, if available:*
There are some CloudStack VMs that does not have assoicated CloudStackMachines in kubernetes.
*Description of changes:*
When CloudStackMachines resource is deleted, there is a possibility that its Spec.InstanceID has not been set due to race condition. This will cause CloudStackMachine resource being deleted without destroy CloudStack VM.

The fix is to return temp error and reconciler will retry and CloudStackMachines'  InstanceID will be set or can be retrieved by using CloudStack API call.
*Testing performed:*
Unit testing, manual testing.


controllers/cloudstackmachine_controller.go
```diff
@@ -268,6 +269,17 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateMachineStateChecker()
 }
 
 func (r *CloudStackMachineReconciliationRunner) ReconcileDelete() (retRes ctrl.Result, reterr error) {
+	if r.ReconciliationSubject.Spec.InstanceID == nil {
+		// InstanceID is not set until deploying VM finishes which can take minutes, and CloudStack Machine can be deleted before VM deployment complete.
+		// ResolveVMInstanceDetails can get InstanceID by CS machine name
+		err := r.CSClient.ResolveVMInstanceDetails(r.ReconciliationSubject)
+		if err != nil {
```
maxdrib, line 276: what if the VM isn't present at all in ACS, if it was never created? We may want to exit then

controllers/cloudstackmachine_controller.go
```diff
@@ -268,6 +269,17 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateMachineStateChecker()
 }
 
 func (r *CloudStackMachineReconciliationRunner) ReconcileDelete() (retRes ctrl.Result, reterr error) {
+	if r.ReconciliationSubject.Spec.InstanceID == nil {
+		// InstanceID is not set until deploying VM finishes which can take minutes, and CloudStack Machine can be deleted before VM deployment complete.
+		// ResolveVMInstanceDetails can get InstanceID by CS machine name
+		err := r.CSClient.ResolveVMInstanceDetails(r.ReconciliationSubject)
```
maxdrib, line 275: does this method automatically set instanceID? Should it return the value, instead of just an error?

pkg/cloud/instance.go
```diff
@@ -282,7 +282,7 @@ func (c *client) GetOrCreateVMInstance(
 		return err
 	}
 	csMachine.Spec.InstanceID = pointer.StringPtr(deployVMResp.Id)
-
+	csMachine.Status.Status = metav1.StatusSuccess
```
maxdrib, line 285: will this cause a panic for older machines which have `csMachine.Status = nil`?

test/dummies/v1beta2/vars.go
```diff
@@ -423,7 +423,7 @@ func SetDummyCAPIMachineVars() {
 		},
 		Spec: clusterv1.MachineSpec{
 			ClusterName:   ClusterName,
-			FailureDomain: pointer.String(Zone1.ID)},
+			FailureDomain: pointer.String("fd1")},
```
maxdrib, line 426: do we not have a constant for this?

wanyufe: We only have couple of places uses this `fd1`. And I changed this zone.id to 'fd1' since we have this multiple-end points feature added.

pkg/cloud/instance.go
wanyufe: csMachine.Status is a struct in previous version, so it cannot be nil.

controllers/cloudstackmachine_controller.go
wanyufe: Yes, ResolveVMInstanceDetails automatically set the instanceID. It is an existed function, we are trying to re-call it to get latest info from CloudStack management API. And instanceID is part of the thing it sets.

controllers/cloudstackmachine_controller.go
wanyufe: This is a tough question - whether we exit or not.
For example, ee got an 'no match found' error
case 1:
This could be a race condition, the VM may in the process of deployment, but not done. The next retry may found it.
case 2:
This VM could be manually deleted through other ways, like CloudStack UI. And this `no match found` could be permanent. Controller enters into rate limited retry loop(up to once every around 16 mins)

Here we follow the best practice:

- set error in status/reason

- Keep retry
Eventually, admin can remove the finalizer, and CloudStackMachine will be GCed by kubernetes runtime. 

A similar situation in Kubernetes is pod having an invalid image configuration, and keeps getting crashloopbackoff until admin fix it.

maxdrib: The pod image configuration is a different situation - it's not a race condition. It will only happen if a cluster is misconfigured by the user, and would only happen once. The operational implications are different. I think we should be more optimistic/automated with this race condition problem and exit on 'no match found' error instead of naively retrying. In a slow environment, I believe we're very likely to have CAPI machines created/deleted without a corresponding ACS VM. I do not think we should require a manual intervention every time this happens, it will be way too much overhead to be sustainable, with create/upgrade/delete operations failing nondeterministically now.

wanyufe: For race-condition situation, we don't have a problem here. The next retry will get the instanceID and cloudstack VM will be deleted and its associated cloudstackmachine finalizer will be removed -- everything is fine here.

Here we are trying to address cloudstackmachine object is deleted while leaving cloudstack VM not being destroyed. I don't know a situation that a cloudstackmachine created without having a cloudstack VM -- if this happens, the controller will keep retrying to create this VM.

Do you have any reason for `I believe we're very likely to have CAPI machines created/deleted without a corresponding ACS VM` ?

wanyufe: I think the challenge is to distinguish above 2 cases I described, even we exit (with/without error), I assume it is case 2 and it is a risky assumption.

controllers/cloudstackmachine_controller.go
```diff
@@ -268,21 +270,33 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateMachineStateChecker()
...
-				r.Log.Info(err.Error())
-				return ctrl.Result{RequeueAfter: utils.DestoryVMRequeueInterval}, nil
-			}
+	if r.ReconciliationSubject.Spec.InstanceID == nil {
+		// InstanceID is not set until deploying VM finishes which can take minutes, and CloudStack Machine can be deleted before VM deployment complete.
+		// ResolveVMInstanceDetails can get InstanceID by CS machine name
+		err := r.CSClient.ResolveVMInstanceDetails(r.ReconciliationSubject)
+		if err != nil {
+			r.ReconciliationSubject.Status.Status = pointer.String(metav1.StatusFailure)
+			r.ReconciliationSubject.Status.Reason = pointer.String(err.Error() +
```
maxdrib, line 279: do you have to defer a Patch operation on this object to get this status update to happen?

controllers/cloudstackmachine_controller.go
maxdrib: My thinking is this - suppose a client creates 1000 CAPI machines all at once. ACS will put those requests into a queue right? Suppose before the MHC timeout, it successfully creates 50 VM’s. Then MHC goes in and tries to delete all the timed out CAPI machines. Doesn’t that leave us in a situation where 950 CAPI machines with no associated VM need to be manually deleted by an admin? My concern is that’s a ton of overhead

controllers/cloudstackmachine_controller.go
wanyufe: ResolveVMInstanceDetails is one of 3 function exposed through VMIface

1. GetOrCreateVMInstance
2. ResolveVMInstanceDetails
3. DestroyVMInstance

Setting InstanceID is only part of its function, and change its func name to indicate that it sets instanceID kind of misleading since it also setting other things. 

I can add a comments before setting the value, so that future developer know its usage when refactoring code.
```
        csMachine.Spec.ProviderID = pointer.StringPtr(fmt.Sprintf("cloudstack:///%s", vmResponse.Id))
	// InstanceID is later used as required parameter to destroy VM.
	csMachine.Spec.InstanceID = pointer.StringPtr(vmResponse.Id)
	
```

controllers/cloudstackmachine_controller.go
wanyufe: Patch operation is part of the base reconciler which runs for all controller.

controllers/cloudstackmachine_controller.go
wanyufe: In this condition, I think no manual overhead.

950 Cloudstackmachine object will be created in etcd, and CAPC controller will try to reconcile and deploy 950 VMs in cloudstack - it takes time, but before it finishes, deleting (caused by MHC) those cloudstackmachine will return error and retry.

DeployVM can fail with an InstanceID which can be later retrieved by ResolveVMInstanceDetails.

maxdrib: What's the termination condition in this scenario? When does the CAPC VM creation workflow terminate? When does this deletion workflow terminate? I'm struggling to see a clean outcome for all the CloudStackMachines being cleanly deleted when failures occur during both VM creation and deletion. Can you explain again? or I'd be happy to discuss over slack

wanyufe: During cloudstackmachine creation: the CAPI machine will be deleted right away (no MHC involve) when VM deployment failed (the instancestate will be error), in this case, we still have a VM instanceID. So the CAPI machine and VM (in error state) will be cleaned up.

During cloudstackmachine deletion: CAPC controller will try to destroy using instanceID, if destroying failed, it will keep retrying. if failed with error message `not found`, it considered success.
CAPC controller will keep retrying to get instanceID when instanceID is missing.

maxdrib (approved):

PR #196, opened 2022-11-30
Multi vm affinity e2e

jweite-amazon:
*Issue #, if available:*

*Description of changes:*
Changes to E2E affinity tests to support true affinity testing if ACS env has three or more hosts in the zone being used for the test.  If fewer than three hosts are available, falls back to the pre-existing one-VM test, which only verifies that the VM is placed in the appropriate affinity group.

*Testing performed:*
Executed test in single host and 3-host environments, observing it allocating the one or three VMs respectively for the CP and workers.  Tests pass in both cases.

### DIFF
--- a/test/e2e/affinity_group.go
+++ b/test/e2e/affinity_group.go
@@ -79,6 +79,17 @@ func AffinityGroupSpec(ctx context.Context, inputGetter func() CommonSpecInput) 
 }
 
 func executeTest(ctx context.Context, input CommonSpecInput, namespace *corev1.Namespace, specName string, clusterResources *clusterctl.ApplyClusterTemplateAndWaitResult, affinityType string) []string {
+	csClient := CreateCloudStackClient(ctx, input.BootstrapClusterProxy.GetKubeconfigPath())
+	zoneName := input.E2EConfig.GetVariable("CLOUDSTACK_ZONE_NAME")
+	numHosts := GetHostCount(csClient, zoneName)
+	machineCount := int64(3) // Must allocate an odd number of CP machines w/ stacked etcd
+	if numHosts < 3 {
+		// If not enough ACS hosts to support 3 machines with anti-affinity, fall back to just allocating one machine.
+		//  Test will only confirm that the single VM is in an anti-affinity group, but not that ACS anti-affinty works.
+		By("Fewer than three host in ACS env.  Only verifying that a single VM gets placed in an AG, not that true affinity/anti-affinity is ultimately achieved.")
+		machineCount = int64(1)
+	}
+
 	clusterctl.ApplyClusterTemplateAndWait(ctx, clusterctl.ApplyClusterTemplateAndWaitInput{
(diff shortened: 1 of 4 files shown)


test/e2e/common.go
```diff
@@ -287,6 +287,24 @@ func CheckAffinityGroupsDeleted(client *cloudstack.CloudStackClient, affinityIds
...
+	Expect(listZonesResponse.Count).To(Equal(1), "multiple zones resolve to zone name %s", zoneName)
+	zoneId := listZonesResponse.Zones[0].Id
+
+	ph := client.Host.NewListHostsParams()
+	ph.SetZoneid(zoneId)
+	ph.SetHypervisor("KVM")
+	ph.SetResourcestate("Enabled")
+	ph.SetState("Up")
+	listHostsResponse, err := client.Host.ListHosts(ph)
+	Expect(err).To(BeNil(), "error listing hosts")
```
maxdrib, line 304: do the contents of `err` get logged here? That'd be useful for troubleshooting.

jweite-amazon: Yes:
`  error listing zones
  Expected
      <*errors.errorString | 0xc00154abe0>: {
          s: "CloudStack API error 401 (CSExceptionErrorCode: 0): unable to verify user credentials and/or request signature",
      }
  to be nil`

test/e2e/affinity_group.go
```diff
@@ -79,6 +79,17 @@ func AffinityGroupSpec(ctx context.Context, inputGetter func() CommonSpecInput)
...
 
 func executeTest(ctx context.Context, input CommonSpecInput, namespace *corev1.Namespace, specName string, clusterResources *clusterctl.ApplyClusterTemplateAndWaitResult, affinityType string) []string {
+	csClient := CreateCloudStackClient(ctx, input.BootstrapClusterProxy.GetKubeconfigPath())
+	zoneName := input.E2EConfig.GetVariable("CLOUDSTACK_ZONE_NAME")
+	numHosts := GetHostCount(csClient, zoneName)
+	machineCount := int64(3) // Must allocate an odd number of CP machines w/ stacked etcd
+	if numHosts < 3 {
+		// If not enough ACS hosts to support 3 machines with anti-affinity, fall back to just allocating one machine.
+		//  Test will only confirm that the single VM is in an anti-affinity group, but not that ACS anti-affinty works.
+		By("Fewer than three host in ACS env.  Only verifying that a single VM gets placed in an AG, not that true affinity/anti-affinity is ultimately achieved.")
```
maxdrib, line 89: I think we should explicitly skip this test if there are <3 hosts. This way is very subtle and tests some plumbing code but not the full vertical feature. If we skip the test with a message explaining why, then the user knows they should address it in their environment.

jweite-amazon: Done. That makes the most sense. Thanks for that Max.

maxdrib (approved):

PR #198, opened 2022-12-05
add events to cloudstack machine resource

wanyufe:
*Issue #, if available:*
For better diagnosis cloudstack machine controller issue, several events are added to CloudStackMachine during reconcile.
*Description of changes:*
Events are created during reconciling CloudStackMachine
*Testing performed:*
Unit testing, manual testing. 

```
Events:
  Type    Reason    Age                    From                     Message
  ----    ------    ----                   ----                     -------
  Normal  Creating  5m30s (x19 over 7m)    capc-machine-controller  Bootstrap DataSecretName not yet available.
  Normal  Created   5m21s                  capc-machine-controller  CloudStack instance Created
  Normal  Running   5m20s (x2 over 5m21s)  capc-machine-controller  Machine instance is Running...
```

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -48,6 +48,19 @@ var (
 	failuredomainMatcher = regexp.MustCompile(`ds\.meta_data\.failuredomain`)
 )
 
+const (
+	BootstrapDataNotReady                      = "Bootstrap DataSecretName not yet available"
+	CSMachineCreationSuccess                   = "CloudStack instance Created"
+	CSMachineCreationFailed                    = "Creating CloudStack machine failed: %s"
+	MachineInstanceRunning                     = "Machine instance is Running..."
+	MachineInErrorMessage                      = "CloudStackMachine VM in error state. Deleting associated Machine"
+	MachineNotReadyMessage                     = "Instance not ready, is %s"
+	CSMachineStateCheckerCreationFailed        = "error encountered when creating CloudStackMachineStateChecker"
+	CSMachineStateCheckerCreationSuccess       = "CloudStackMachineStateChecker created"
+	CSMachineDeletionMessage                   = "Deleting CloudStack Machine %s"
+	CSMachineDeletionInstanceIDNotFoundMessage = "Deleting CloudStack Machine %s instanceID not found"
+)
+
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=cloudstackmachines,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=cloudstackmachines/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=cloudstackmachines/finalizers,verbs=update
@@ -187,8 +200,8 @@ func (r *CloudStackMachineReconciliationRunner) DeleteMachineIfFailuredomainNotE
 // Implicitly it also fetches its bootstrap secret in order to create said instance.
 func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes ctrl.Result, reterr error) {
 	if r.CAPIMachine.Spec.Bootstrap.DataSecretName == nil {
-		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Creating", "Bootstrap DataSecretName not yet available")
-		return r.RequeueWithMessage("Bootstrap DataSecretName not yet available.")
+		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Creating", BootstrapDataNotReady)
+		return r.RequeueWithMessage(BootstrapDataNotReady + ".")
 	}
 	r.Log.Info("Got Bootstrap DataSecretName.")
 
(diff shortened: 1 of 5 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -186,6 +187,7 @@ func (r *CloudStackMachineReconciliationRunner) DeleteMachineIfFailuredomainNotE
 // Implicitly it also fetches its bootstrap secret in order to create said instance.
 func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes ctrl.Result, reterr error) {
 	if r.CAPIMachine.Spec.Bootstrap.DataSecretName == nil {
+		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Creating", "Bootstrap DataSecretName not yet available.")
 		return r.RequeueWithMessage("Bootstrap DataSecretName not yet available.")
```
maxdrib, line 191: should we instead refactor RequeueWithMessage to always write an event to the Reconciliation subject?

controllers/cloudstackmachine_controller.go
```diff
@@ -186,6 +187,7 @@ func (r *CloudStackMachineReconciliationRunner) DeleteMachineIfFailuredomainNotE
 // Implicitly it also fetches its bootstrap secret in order to create said instance.
 func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes ctrl.Result, reterr error) {
 	if r.CAPIMachine.Spec.Bootstrap.DataSecretName == nil {
+		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Creating", "Bootstrap DataSecretName not yet available.")
```
maxdrib, line 190: why not write to the CAPI machine?

controllers/cloudstackmachine_controller.go
```diff
@@ -223,15 +229,18 @@ func processCustomMetadata(data []byte, r *CloudStackMachineReconciliationRunner
...
 		r.ReconciliationSubject.Status.Ready = true
 	} else if r.ReconciliationSubject.Status.InstanceState == "Error" {
+		r.Recorder.Event(r.ReconciliationSubject, "Warning", "Error", "CloudStackMachine VM in error state. Deleting associated Machine.")
 		r.Log.Info("CloudStackMachine VM in error state. Deleting associated Machine.", "csMachine", r.ReconciliationSubject.GetName())
 		if err := r.K8sClient.Delete(r.RequestCtx, r.CAPIMachine); err != nil {
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: utils.RequeueTimeout}, nil
 	} else {
+		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", r.ReconciliationSubject.Status.InstanceState, "Instance not ready, is %s.", r.ReconciliationSubject.Status.InstanceState)
```
maxdrib, line 243: Nit: Could you be consistent with punctuation? I see periods somewhere but not everywhere

controllers/cloudstackmachine_controller.go
wanyufe: requeueWithMessage can potentially be used by other controller (not capc machine controller), it may make the scope of this PR. When other controllers use requeueWithMessage, it may need attach events to different resources.

controllers/cloudstackmachine_controller.go
wanyufe: Tried in my local, technically, we can create an event and bind it to CAPI machine. 
After discussed with team, we concluded that putting events under CAPC only was enough right now. 

Thanks for the idea.

controllers/cloudstackmachine_controller.go
```diff
@@ -54,6 +54,7 @@ var (
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets,verbs=get;list;watch
 // +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=kubeadmcontrolplanes,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```
maxdrib, line 57: should we restrict this to only the capc group?

controllers/cloudstackmachine_controller.go
```diff
@@ -186,6 +187,7 @@ func (r *CloudStackMachineReconciliationRunner) DeleteMachineIfFailuredomainNotE
 // Implicitly it also fetches its bootstrap secret in order to create said instance.
 func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes ctrl.Result, reterr error) {
 	if r.CAPIMachine.Spec.Bootstrap.DataSecretName == nil {
+		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Creating", "Bootstrap DataSecretName not yet available")
```
maxdrib, line 190: we should reuse the log strings in these messages to prevent divergence in the future. Same below.

controllers/cloudstackmachine_controller.go
```diff
@@ -223,15 +229,18 @@ func processCustomMetadata(data []byte, r *CloudStackMachineReconciliationRunner
...
 func (r *CloudStackMachineReconciliationRunner) RequeueIfInstanceNotRunning() (retRes ctrl.Result, reterr error) {
 	if r.ReconciliationSubject.Status.InstanceState == "Running" {
+		r.Recorder.Event(r.ReconciliationSubject, "Normal", "Running", "Machine instance is Running...")
 		r.Log.Info("Machine instance is Running...")
 		r.ReconciliationSubject.Status.Ready = true
 	} else if r.ReconciliationSubject.Status.InstanceState == "Error" {
+		r.Recorder.Event(r.ReconciliationSubject, "Warning", "Error", "CloudStackMachine VM in error state. Deleting associated Machine")
 		r.Log.Info("CloudStackMachine VM in error state. Deleting associated Machine.", "csMachine", r.ReconciliationSubject.GetName())
 		if err := r.K8sClient.Delete(r.RequestCtx, r.CAPIMachine); err != nil {
 			return ctrl.Result{}, err
```
maxdrib, line 239: should we capture this error as an event as well?

maxdrib (commented):
Minor comments - overall logic looks good to me

controllers/cloudstackmachine_controller.go
maxdrib: Ah never mind, I thought this call to `Delete` was to CloudStack, not the K8sClient. I think we can disregard.

controllers/cloudstackmachine_controller.go
wanyufe: The group of event is "" (empty string), means it is inside core like `pod`
If we change it to `// +kubebuilder:rbac:groups="**infrastructure.cluster.x-k8s.io**",resources=events,verbs=create;patch`, resource events is not a valid value inside that groups.

controllers/cloudstackmachine_controller.go
wanyufe: refactor to use cons string for logs and events.

PR #200, opened 2022-12-07
Add exceeding account quota e2e test case

wongni:
*Issue #, if available:*

*Description of changes:*

Adding an e2e test case that test exceeding account's quota (e.g. instance limits)

*Testing performed:*

e2e testing


maxdrib (approved):
/lgtm

davidjumani:
IMO this test is not needed as it tests a feature in CloudStack rather than CAPC and there are already existing tests in CloudStack to test quotas

chrisdoherty4:
> IMO this test is not needed as it tests against a feature in CloudStack and just specific error handling in CAPC (deployments would fail nonetheless). There are already existing tests in CloudStack to test quotas

I'm inclined to agree. If the error were bubbled up to the API object in some capacity (perhaps a condition) then we might want a test; that's not happening right now so this seems superfluous.

jweite-amazon:
I'm fine with closing this.

chrisdoherty4:
@jweite-amazon thanks!

rohityadavcloud:
Closing this on @jweite-amazon 's remarks. Thanks.

PR #201, opened 2022-12-07
Add finalizer only when deployVM no error

wanyufe:
*Issue #, if available:*
When deployVM failed, for example, due to invalid serviceOffering. The instanceID is not set and CAPC object could not be deleted due to missing InstanceID.
*Description of changes:*
Do not add finalizer to CAPC object when error happened during deployVM.
*Testing performed:*
Manual testing.

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -223,11 +223,11 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes 
 		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", "Creating", CSMachineCreationFailed, err.Error())
 	}
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
+		controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
 		r.Recorder.Eventf(r.ReconciliationSubject, "Normal", "Created", CSMachineCreationSuccess)
 		r.Log.Info(CSMachineCreationSuccess, "instanceStatus", r.ReconciliationSubject.Status)
 	}
-	// Always add the finalizer regardless. It can't be added twice anyway.
-	controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
+	
 	return ctrl.Result{}, err
 }
 
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -223,11 +223,14 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes 
 		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", "Creating", CSMachineCreationFailed, err.Error())
 	}
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
+		// adding a finalizer will require CPAC reconcile-delete trying to destroy associated VM through instanceID,
+		// so if err is not nil, which means it could not get associated VM through instanceID or name, we should not add finalizer to CloudStackMachine,
+		// otherwise, reconciler-delete will be stuck trying to wait instanceID to ba available.
 		controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
 		r.Recorder.Eventf(r.ReconciliationSubject, "Normal", "Created", CSMachineCreationSuccess)
 		r.Log.Info(CSMachineCreationSuccess, "instanceStatus", r.ReconciliationSubject.Status)
 	}
-	
+
 	return ctrl.Result{}, err
 }
 
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -223,9 +223,9 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes 
 		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", "Creating", CSMachineCreationFailed, err.Error())
 	}
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
-		// adding a finalizer will require CPAC reconcile-delete trying to destroy associated VM through instanceID,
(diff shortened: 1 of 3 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -223,11 +223,11 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes
 		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", "Creating", CSMachineCreationFailed, err.Error())
 	}
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
+		controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
 		r.Recorder.Eventf(r.ReconciliationSubject, "Normal", "Created", CSMachineCreationSuccess)
 		r.Log.Info(CSMachineCreationSuccess, "instanceStatus", r.ReconciliationSubject.Status)
 	}
-	// Always add the finalizer regardless. It can't be added twice anyway.
-	controllerutil.AddFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer)
```
maxdrib, line 230: would there ever be any resources under the cloudstackmachine which would need to be cleaned up? Another layer of abstraction? What if we start managing volumes independently of machines someday? Wouldn't a lack of a finalizer result in orphaned sub-resources? It's not a blocker, but I want to be sure we're confident in this approach

pkg/cloud/instance.go
```diff
@@ -277,13 +277,15 @@ func (c *client) GetOrCreateVMInstance(
 		listVirtualMachineParams.SetName(csMachine.Name)
 		if listVirtualMachinesResponse, err2 := c.cs.VirtualMachine.ListVirtualMachines(listVirtualMachineParams); err2 == nil && listVirtualMachinesResponse.Count > 0 {
 			csMachine.Spec.InstanceID = pointer.StringPtr(listVirtualMachinesResponse.VirtualMachines[0].Id)
+			csMachine.Status.InstanceState = listVirtualMachinesResponse.VirtualMachines[0].State
 		} else {
 			c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err2)
+			return err
 		}
```
maxdrib, line 284: I prefer avoiding `else` statements wherever possible :) https://github.com/aws/eks-anywhere/blob/main/docs/developer/best-practice.md#returns

```suggestion
		if listVirtualMachinesResponse, err2 := c.cs.VirtualMachine.ListVirtualMachines(listVirtualMachineParams); err2 != nil || listVirtualMachinesResponse.Count == 0 {
			c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err2)
			return err
		}
		csMachine.Spec.InstanceID = pointer.StringPtr(listVirtualMachinesResponse.VirtualMachines[0].Id)
		csMachine.Status.InstanceState = listVirtualMachinesResponse.VirtualMachines[0].State
```

controllers/cloudstackmachine_controller.go
jweite-amazon: I suspect it's impossible to predict exactly how a future developer will modify things.  If code to create the proposed sub-resources runs after VM creation, then I believe these will never be created, so all is well.  If such code is added before VM creation, then yes, these would be created, and could be orphaned at delete time.  Since this is non-obvious behavior, perhaps best we can do is add a highly visible comment (which should also mention the problem that occurs if the finalizer is created despite VM creation failure).

wanyufe: If some other resource owned (created ) by cloudstackmachine, that owned resource will have an ownerRef to cloudstackmachine. Once cloudstackMachine is deleted, owned resource will be deleted too by kube -- there are several variants of this depends on deletion cascade policy.

pkg/cloud/instance.go
wanyufe: This is a good suggestion. But seems not working here. 
the `listVirtualMachinesResponse` is defined in if statement, and could not be used later on -- IDE is complaining `listVirtualMachinesResponse` is not defined in line 282

controllers/cloudstackmachine_controller.go
jweite-amazon: Fair enough.  I'd still like to see a comment to inform future devs about the reason for and implications of the lifecycle management technique coded.

pkg/cloud/instance.go
```diff
@@ -275,15 +275,17 @@ func (c *client) GetOrCreateVMInstance(
...
-		} else {
+		listVirtualMachinesResponse, err2 := c.cs.VirtualMachine.ListVirtualMachines(listVirtualMachineParams)
+		if err2 != nil || listVirtualMachinesResponse.Count <= 0 {
 			c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err2)
+			return err
 		}
-		return err
+		csMachine.Spec.InstanceID = pointer.StringPtr(listVirtualMachinesResponse.VirtualMachines[0].Id)
+		csMachine.Status.InstanceState = listVirtualMachinesResponse.VirtualMachines[0].State
+	} else {
```
jweite-amazon, line 285: I think the rationale for lifting the scope of deployVMResp out of the if was so that this else could be eliminated (i.e., the code within it be move into the main-line flow).

controllers/cloudstackmachine_controller.go
```diff
@@ -223,11 +223,14 @@ func (r *CloudStackMachineReconciliationRunner) GetOrCreateVMInstance() (retRes
 		r.Recorder.Eventf(r.ReconciliationSubject, "Warning", "Creating", CSMachineCreationFailed, err.Error())
 	}
 	if err == nil && !controllerutil.ContainsFinalizer(r.ReconciliationSubject, infrav1.MachineFinalizer) { // Fetched or Created?
+		// adding a finalizer will require CPAC reconcile-delete trying to destroy associated VM through instanceID,
+		// so if err is not nil, which means it could not get associated VM through instanceID or name, we should not add finalizer to CloudStackMachine,
+		// otherwise, reconciler-delete will be stuck trying to wait instanceID to ba available.
```
maxdrib, line 228:
```suggestion
		// Adding a finalizer will make reconcile-delete try to destroy the associated VM through its instanceID.
		// If err is not nil, it means CAPC could not get an associated VM through instanceID or name, so we should not add a finalizer to this CloudStackMachine.
		// Otherwise, reconcile-delete will be stuck trying to wait for instanceID to be available.
```

pkg/cloud/instance.go
jweite-amazon: My bad here, I misread the indentation in the github diff.  Understood now that the else block I called out is not the one Max asked be removed, and that it need to stand.

maxdrib (approved):

Issue #207, opened 2023-01-05
CAPC does not resolve service offering with zone as additional filtering parameter


chrsmeca:
/kind bug

**What steps did you take and what happened:**
The following error is generated when multiple cloud stack zones with the same offering name exist.

E0104 21:43:15.451029       1 controller.go:317] controller/cloudstackmachine "msg"="Reconciler error" "error"="1 error occurred:\n\t* expected 1 Service Offering with name 4 vCPU 8 GiB RAM, but got 2\n\n" "name"="<<redacted>>" "namespace"="<<redacted>>" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="CloudStackMachine" 

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-cloudstack version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

wanyufe:
PR https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/208 fixed the issue.

PR #208, opened 2023-01-05
Add zoneID parameter when retrieving service/disk offering

wanyufe:
*Issue #, if available:*
[CAPC does not resolve service offering with zone as additional filtering parameter](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/207)
Also noticed similar issue happened to diskoffering too.
*Description of changes:*
Add zoneID when calling CloudStack API to retrieve service offering by name or retrieve disk offering by name
*Testing performed:*
Unit Test, manual testing.

### DIFF
--- a/pkg/cloud/instance.go
+++ b/pkg/cloud/instance.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -84,7 +84,7 @@ func (c *client) ResolveVMInstanceDetails(csMachine *infrav1.CloudStackMachine) 
 	return errors.New("no match found")
 }
 
-func (c *client) ResolveServiceOffering(csMachine *infrav1.CloudStackMachine) (offeringID string, retErr error) {
+func (c *client) ResolveServiceOffering(csMachine *infrav1.CloudStackMachine, zoneID string) (offeringID string, retErr error) {
 	if len(csMachine.Spec.Offering.ID) > 0 {
 		csOffering, count, err := c.cs.ServiceOffering.GetServiceOfferingByID(csMachine.Spec.Offering.ID)
 		if err != nil {
@@ -102,14 +102,14 @@ func (c *client) ResolveServiceOffering(csMachine *infrav1.CloudStackMachine) (o
 		}
 		return csMachine.Spec.Offering.ID, nil
 	}
-	offeringID, count, err := c.cs.ServiceOffering.GetServiceOfferingID(csMachine.Spec.Offering.Name)
+	offeringID, count, err := c.cs.ServiceOffering.GetServiceOfferingID(csMachine.Spec.Offering.Name, cloudstack.WithZone(zoneID))
 	if err != nil {
 		c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err)
 		return "", multierror.Append(retErr, errors.Wrapf(
-			err, "could not get Service Offering ID from %s", csMachine.Spec.Offering.Name))
+			err, "could not get Service Offering ID from %s in zone %s", csMachine.Spec.Offering.Name, zoneID))
 	} else if count != 1 {
 		return "", multierror.Append(retErr, errors.Errorf(
-			"expected 1 Service Offering with name %s, but got %d", csMachine.Spec.Offering.Name, count))
+			"expected 1 Service Offering with name %s in zone %s, but got %d", csMachine.Spec.Offering.Name, zoneID, count))
 	}
 	return offeringID, nil
 }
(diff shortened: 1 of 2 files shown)


wanyufe:
#207

api/v1beta2/cloudstackaffinitygroup_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
wongni, line 2: This is not updated at all. I think you can keep 2022.

pkg/cloud/instance.go
```diff
@@ -102,14 +102,14 @@ func (c *client) ResolveServiceOffering(csMachine *infrav1.CloudStackMachine) (o
 		}
 		return csMachine.Spec.Offering.ID, nil
 	}
-	offeringID, count, err := c.cs.ServiceOffering.GetServiceOfferingID(csMachine.Spec.Offering.Name)
+	offeringID, count, err := c.cs.ServiceOffering.GetServiceOfferingID(csMachine.Spec.Offering.Name, cloudstack.WithZone(zoneID))
```
wongni, line 105: Nice on using withZone function instead of listing service offerings!

pkg/cloud/instance_test.go
```diff
@@ -143,23 +143,23 @@ var _ = Describe("Instance", func() {
 
 		It("returns errors occurring while fetching service offering information", func() {
 			expectVMNotFound()
-			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering.Name).Return("", -1, unknownError)
+			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering.Name, gomock.Any()).Return("", -1, unknownError)
```
wongni, line 146:
```suggestion
			sos.EXPECT().GetServiceOfferingID(dummies.CSMachine1.Spec.Offering.Name, cloudstack.WithZone(dummies.XXX.zoneID)).Return("", -1, unknownError)
```
Why not expecting a real value instead of Any()?

pkg/metrics/metrics.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
wongni, line 2: This file is not updated.

pkg/webhookutil/webhook_validators.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
wongni, line 2: This file is not updated.

pkg/cloud/instance_test.go
wanyufe: We don't care the value here, so we use gomock.Any to refer it. 

If we use `cloudstack.WithZone(...`, we will get the follwowing error: 
`
 Got: [0x229b1c0] ([]interface {})
  Want: is equal to 0x229b1c0 (cloudstack.OptionFunc)
`

I am sure there is a work-around on above error, since we are using that value, I used gomock.Any() to refer it.

pkg/metrics/metrics.go
wanyufe: There are too many files needs to be changed. So create another PR to address it separately: 
https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/209

api/v1beta2/cloudstackaffinitygroup_types.go
wanyufe: I keep getting 
`Error: pkg/webhookutil/webhook_validators.go:1:11: Expected:2023, Actual: 2022 The Kubernetes Authors. (goheader)` and similar error even this PR does not change those files. 

I created a PR to https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/209 to address above error.

pkg/webhookutil/webhook_validators.go
wanyufe: Similar to above explanation.

api/v1beta2/cloudstackaffinitygroup_types.go
mrog: Would it work if you change it to `Copyright 2022-2023`?  That would be more accurate.

pkg/cloud/instance.go
wanyufe: it uses list service offering internally, just seteting zoneID as additional parameter before calling.

api/v1beta2/cloudstackaffinitygroup_types.go
wanyufe: I answered this question on PR https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/209

wongni (approved):

davidjumani:
CLGTM
/lgtm
/approve

PR #209, opened 2023-01-05
Update copyright header from 2022 to 2023

wanyufe:
*Issue #, if available:*
Copyright header **2022** causing build warning.
*Description of changes:*
Update *.go copyright header
*Testing performed:*
Unit testing

### DIFF
--- a/api/v1beta1/cloudstackaffinitygroup_conversion.go
+++ b/api/v1beta1/cloudstackaffinitygroup_conversion.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/api/v1beta1/cloudstackaffinitygroup_types.go
+++ b/api/v1beta1/cloudstackaffinitygroup_types.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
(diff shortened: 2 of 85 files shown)


api/v1beta1/cloudstackaffinitygroup_conversion.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
mrog, line 2: Would it be better to change it to `2022-2023`?

wanyufe: I honest don't know. This maybe a legal question. I did not see 2021-2022 in vsphere .go files even we know it started before 2022.

mrog: I looked at some of the vsphere files, and they contain the year that the file was created, not the year it was last updated.  Maybe we should leave the copyrights alone?

wanyufe: I agree with you on this. 
But our repo has a `goheader` check that tries to make sure the copyright year is current year. Maybe we should not include it in our lint checking list. 

@davidjumani @jweite-amazon Do you have any opinion on this?

davidjumani: Looking at other projects, it is the year created, not the current year. IMO the linter should be changed

wanyufe:
close, create a new solution: https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/211

PR #221, opened 2023-02-21
GetPublicIPs: never choose an already assigned IP

vdombrovski:
*Description of changes:*

This fix removes a condition where an already assigned public IP inside the same Isolated network would be chosen as the IP of another cluster. This condition could only be triggered whenever a new controlPlaneEndpoint is chosen for a new cluster.

*Testing performed:*

Deployed on our staging cluster. Provisioned several clusters inside the same guest network, checked all clusters were deployed properly.

### DIFF
--- a/pkg/cloud/isolated_network.go
+++ b/pkg/cloud/isolated_network.go
@@ -141,12 +141,6 @@ func (c *client) GetPublicIP(
 		// Ignore already allocated here since the IP was specified.
 		return publicAddresses.PublicIpAddresses[0], nil
 	} else if publicAddresses.Count > 0 { // Endpoint not specified.
-		// TODO: Use tags to prevent clash with this logic.
-		for _, v := range publicAddresses.PublicIpAddresses { // Pick first available address.
-			if v.Allocated != "" && v.Associatednetworkid == isoNet.Spec.ID && !v.Issourcenat { // IP Already allocated.
-				return v, nil
-			}
-		}
 		for _, v := range publicAddresses.PublicIpAddresses { // Pick first available address.
 			if v.Allocated == "" { // Found un-allocated Public IP.
 				return v, nil


vdombrovski:
@cla-bot check

jweite-amazon:
@rejoshed, the block of code this submitter is proposing to delete were added during the major CRD refactoring of [PR-55](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/55).  Public IP Address selection logic previously lived in network.go (line 226) and did not include this logic.  While I know you're no longer actively contributing, do you recall the intent behind adding this code?  The _ToDo_ comment about introducing additional tag-based logic to avoid _clashes_ seems to indicate that this was more than exploratory code inadvertantly retained.

vdombrovski:
@jweite-amazon from what I understand, there are 3 possible states the IP could be in:
1. There is a controlPlaneEndpoint defined (by the user or patched by CAPC): this IP will be returned
2. There is no IP defined by the user: the first non-allocated IP will be allocated and returned
_3. CAPC has not yet had time to patch the controlPlaneEndpoint, and the function is invoked. Then the first non source-NAT (aka VirtualRouter) public IP that is allocated inside the network is returned_

This 3rd condition is the one causing issues, because it is fragile and depends on the fact that there is only one cluster per isolated network. Instead, my PR is proposing to return an error, which will simple trigger a retry, giving enough time to CAPC to patch the host, after which, the 1st case will be returned as intended.

mrog:
@vdombrovski I understand states 1 and 2, but I need some help understanding state 3.  Can you please explain the significance of patching the controlPlaneEndpoint, and how that might relate to the IP address?

vdombrovski:
@mrog Let me show you, please take a look at the short reconciliation function for the Isolated Network:

https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/4a4752250feee59be9bec56adf4d5797092aed09/controllers/cloudstackisolatednetwork_controller.go#L69

a. GetOrCreateIsolatedNetwork is invoked, which assigns a public IP (if not assigned already), adds a firewall rule + load-balancer rule on target apiserver port = multiple ACS API requests
b. AddClusterTagadds the tags to the entities (e.g. created_by_capc, CAPC_cluster_XYZ...) = ACS API requests
c. Patch finally patches the endpoint back into the spec = APIServer request on the host cluster

Between a. and c. there is a delay t>0, which opens opportunity for a race condition (maybe the CAPC pod crashes before c., then his successor restarts the reconciliation). With my fix, this would cause another public IP to be associated instead, which possibly isn't something the original author intended, which lead to him implementing state 3.

However, the way it's implemented leaves opportunity to stealing another associated IP inside the same network, which can cause serious issues (we've had a major incident because of this already). This fix is my attempt at removing the very dangerous condition, while still leaving the highly unlikely race condition in place, which would simply result in an unused public IP. 

Doing this "cleanly" is actually quite hard, because it requires ACS to be aware of which cluster the endpoint belongs to. The original author has suggested using tags, but tagging can not be done atomically when associating an IP address, which, again, leaves opportunity for unlikely race conditions to occur. This is why I haven't attempted the fix described in the TODO, as it wouldn't add any more resilience to the current state of the program.

mrog:
@vdombrovski Thank you.  That explanation helped a lot.  I agree with your assessment.

jweite-amazon:
Thank your for that elaboration @vdombrovski.  Makes sense to me.

/approved

PR #226, opened 2023-03-08
Download golangci-lint instead of building it

hrak:
*Description of changes:*

This PR downloads golangci-lint instead of building it, as recommended by the authors of golangci-lint, and similar to [how its done in CAPI](https://github.com/kubernetes-sigs/cluster-api/pull/4700)

The PR also bumps the version of golangci-lint from v1.46.2 to v1.51.2 and staticcheck from v0.3.1 to 2023.1.2 (v0.4.2)

This will also eventually be required for upgrading the clusterapi dependency, because while testing with clusterapi 1.2.11 i had to update to minimal go 1.19, which required a golangci-lint update, and ran into [the issue outlined here ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4101).

*Testing performed:*

`make build`


hack/tools/tools.go
```diff
@@ -21,7 +21,6 @@ package tools
 import (
 	_ "github.com/a8m/envsubst"
 	_ "github.com/golang/mock/mockgen"
-	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
```
chrisdoherty4, line 24: If I'm not mistaken, tools like dependabot will struggle to flag anything not in the manifest.

Do you have any insight?

hrak:
A little more on the issue this is trying to solve: https://github.com/ryancurrah/gomodguard/issues/33

Another way of working around the issue is [shown here](https://github.com/ryancurrah/gomodguard/issues/33) where they've changed the build tag for the tools to be something other than `tools`

hack/tools/tools.go
chrisdoherty4: Ah, that's painful. Thanks!

chrisdoherty4 (approved):

Issue #220, opened 2023-02-20
Allow UDP and ICMP egress from isolated network


hrak:
/kind feature

**Describe the solution you'd like**

Right now upon network creation, [the egress rule created by default only allows TCP traffic out.](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/c3f25a2e38d57cc32eeeeee3ceddfc44160444b3/pkg/cloud/isolated_network.go#L115) This blocks things like f.e. NTP, so your Kubernetes nodes are unable to sync with an external NTP pool (amongst other issues).

(Optionally?) allow UDP egress in Cloudstack firewall to resolve this.

hrak:
What are your opinions on this? I would like to know if you agree that egress firewall should default to allow all (TCP/UDP/ICMP). If so i will create a PR.

weizhouapache:
@hrak 
+1

hrak:
I have the patch ready, just need to file a PR for it. Will do ASAP.

davidjumani:
Sorry for not getting back to you sooner, but as mentioned in the PR, I think this should not be added to the CAPC code because this is not within the functionality of CAPI - It should just manage the cluster lifecycle, and not handle additional tasks such as ensuring NTP sync or ICMP works. The user should open this up or any application which integrates with CAPC.

davidjumani:
> @davidjumani i don't quite understand this. If it is not within the functionality of CAPI to allow UDP or ICMP egress, then why is it in the functionality to allow TCP egress?

It is to allow the kubernetes endpoint to communicate with the outside world when queried (either by CAPI or the end user). Since it allows all traffic, rather than just 6443, I think it should be restricted to just this. I was able to bring up a cluster with just this port open. There might be cases where the container images or CNI images might not be present on the node, which is why I think a blanket TCP rule was added to allow for downloading these images.
I think its safer to allow only the necessary traffic (6443) and let the user manually punch holes in the firewall for the application / ntp needs (as this relates to the applications / host OS and not kubernetes operations)

However, since these are egress rules, I don't have any hard opinion on this and will support the decision the team takes :)

chrisdoherty4:
I don't see an issue with opening up egress for these protocols. 

[...]

Issue #228, opened 2023-03-16
Merge the AWS build container into the project


chrisdoherty4:
The CI is failing on PRs often because the Go version is outdated in the build container. When AWS donated CAPC to the CNCF the build container sources we're left behind. This ticket is to track their migration.

[Failing CI example](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/224)
[CI jobs referencing build image](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-cloudstack/capi-provider-cloudstack-presumbit.yaml#L68)

/kind feature
/assign chrisdoherty4

g-gaston:
@chrisdoherty4 now that we use the community maintained image, this is done right?

chrisdoherty4:
Correct, we don't need this.

PR #235, opened 2023-05-05
Open egress firewall for UDP and ICMP too

hrak:
*Issue #, if available:*

#220 

*Description of changes:*

This PR opens the egress firewall for UDP and ICMP too. Currently it only allows TCP traffic to flow out of the cluster, blocking things like NTP etc.

*Testing performed:*

Running for several weeks in testing cluster. Bootstrapped several times in dev cluster.

### DIFF
--- a/docs/book/src/clustercloudstack/configuration.md
+++ b/docs/book/src/clustercloudstack/configuration.md
@@ -61,7 +61,8 @@ cmk list zones listall=true | jq '.zone[] | {name, id}'
 
 The network must be declared as an environment variable `CLOUDSTACK_NETWORK_NAME` and is a mandatory parameter.
 As of now, only isolated and shared networks are supported.
-If the specified network does not exist, a new isolated network will be created.
+
+If the specified network does not exist, a new isolated network will be created. The newly created network will have a default egress firewall policy that allows all TCP, UDP and ICMP traffic from the cluster to the outside world.
 
 The list of networks for the specific zone can be fetched using the cmk cli as follows :
 ```
--- a/pkg/cloud/isolated_network.go
+++ b/pkg/cloud/isolated_network.go
@@ -110,12 +110,26 @@ func (c *client) CreateIsolatedNetwork(fd *infrav1.CloudStackFailureDomain, isoN
 	return c.AddCreatedByCAPCTag(ResourceTypeNetwork, isoNet.Spec.ID)
 }
 
-// OpenFirewallRules opens a CloudStack firewall for an isolated network.
+// OpenFirewallRules opens a CloudStack egress firewall for an isolated network.
 func (c *client) OpenFirewallRules(isoNet *infrav1.CloudStackIsolatedNetwork) (retErr error) {
-	p := c.cs.Firewall.NewCreateEgressFirewallRuleParams(isoNet.Spec.ID, NetworkProtocolTCP)
-	_, retErr = c.cs.Firewall.CreateEgressFirewallRule(p)
-	if retErr != nil && strings.Contains(strings.ToLower(retErr.Error()), "there is already") { // Already a firewall rule here.
-		retErr = nil
+	protocols := []string{NetworkProtocolTCP, NetworkProtocolUDP, NetworkProtocolICMP}
+	for _, proto := range protocols {
+		p := c.cs.Firewall.NewCreateEgressFirewallRuleParams(isoNet.Spec.ID, proto)
+
+		if proto == "icmp" {
+			p.SetIcmptype(-1)
+			p.SetIcmpcode(-1)
+		}
+
+		_, err := c.cs.Firewall.CreateEgressFirewallRule(p)
+		if err != nil &&
+			// Ignore errors regarding already existing fw rules for TCP/UDP
+			!strings.Contains(strings.ToLower(err.Error()), "there is already") &&
+			// Ignore errors regarding already existing fw rule for ICMP
+			!strings.Contains(strings.ToLower(err.Error()), "new rule conflicts with existing rule") {
+			retErr = multierror.Append(retErr, errors.Wrapf(
+				err, "failed creating egress firewall rule for network ID %s protocol %s", isoNet.Spec.ID, proto))
+		}
 	}
 	c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(retErr)
 	return retErr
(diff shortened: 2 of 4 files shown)


weizhouapache:
@hrak 
there are some unit test failures

hrak:
> @hrak there are some unit test failures

@weizhouapache @rohityadavcloud tests have been failing for a while now. Its because the CI container has Go 1.17 and needs 1.19. Fixing it depends on #228

hrak:
the tests are not going to pass here @rohityadavcloud - see earlier comment

rohityadavcloud:
Let's discuss with other stakeholders and build consensus.

hrak:
> @hrak the test failures seem to be related to this PR

Sorry, my bad. Just pushed some fixes.

pkg/cloud/isolated_network.go
```diff
@@ -110,12 +110,24 @@ func (c *client) CreateIsolatedNetwork(fd *infrav1.CloudStackFailureDomain, isoN
...
+	protocols := []string{NetworkProtocolTCP, NetworkProtocolUDP, NetworkProtocolICMP}
+	for _, proto := range protocols {
+		p := c.cs.Firewall.NewCreateEgressFirewallRuleParams(isoNet.Spec.ID, proto)
+
+		if proto == "icmp" {
+			p.SetIcmptype(-1)
+			p.SetIcmpcode(-1)
+		}
+
+		_, retErr = c.cs.Firewall.CreateEgressFirewallRule(p)
```
chrisdoherty4, line 124: Do we want to collect the errors into 1 and return that? (We already depend on hashicorps multi error package).

hrak: Something like this?

```patch
--- a/pkg/cloud/isolated_network.go
+++ b/pkg/cloud/isolated_network.go
@@ -121,12 +121,10 @@ func (c *client) OpenFirewallRules(isoNet *infrav1.CloudStackIsolatedNetwork) (r
 			p.SetIcmpcode(-1)
 		}

-		_, retErr = c.cs.Firewall.CreateEgressFirewallRule(p)
-		if retErr != nil && strings.Contains(strings.ToLower(retErr.Error()), "there is already") { // Already a firewall rule here.
-			retErr = nil
-		}
-		if retErr != nil {
-			break
+		_, err := c.cs.Firewall.CreateEgressFirewallRule(p)
+		if err != nil && !strings.Contains(strings.ToLower(err.Error()), "there is already") { // Already a firewall rule here.
+			multierror.Append(retErr, errors.Wrapf(
+				err, "failed creating egress firewall rule for network ID %s protocol %s", isoNet.Spec.ID, proto))
 		}
 	}
 	c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(retErr)
 ```

Comment (chrisdoherty4) [reply pkg/cloud/isolated_network.go:124]: Yes, that looks good.

**Comment by chrisdoherty4** [comment]
@g-gaston

**Comment by g-gaston** [comment]
I have seen this mentioned as pre-requirement to bump to capi 1.4
But looking at the code, I don't see how those two things are linked.
@hrak, could you clarify on that? It's possible that I misinterpreted it 😅 

Regardless of that, I lack enough context to make a decision on this PR by myself.
Assuming that is not required and given this is a change in behavior that makes the network more open, could we make this optional? Like allowing users to open more or less things through the the capc API.

**Comment by hrak** [comment]
Sorry, that last multierror change broke the lint checks. Just pushed a fix.

@g-gaston its not a requirement for capi 1.4, it would just be nice to get this into the upcoming 0.4.x release with capi 1.2.12, and after that start working on 0.5.x with capi 1.4.

**Comment by g-gaston** [comment]
@weizhouapache @rohityadavcloud @davidjumani

Opinions on making this a default vs making it optional?

I lack context on why the egress was restricted to TCP originally.

**Comment by weizhouapache** [comment]
> @weizhouapache @rohityadavcloud @davidjumani
> 
> Opinions on making this a default vs making it optional?
> 
> I lack context on why the egress was restricted to TCP originally.

@g-gaston @hrak 

In CloudStack, the default network offering for CloudStack Kubernetes Cluster (aka CKS) has defaultegresspolicy=true, which means all outgoing traffic are allowed (Including TCP, UDP, ICMP and other protocols).
(cc@hrak 'all' is a valid protocol in cloudstack)

However, if the network already exists, no egress rules will be added. (User has to allow outoging traffic before creating a CKS cluster)

Ideally, we should check the network offering and add egress rules only if defaultegresspolicy=false. this is related to #206

**Comment by davidjumani** [comment]
> @weizhouapache @rohityadavcloud @davidjumani
> 
> Opinions on making this a default vs making it optional?
> 
> I lack context on why the egress was restricted to TCP originally.

This should be handled by the user entirely. CAPI is designed to get a cluster running and to manage its lifecycle, leaving the rest up to the user. That's the reason only TCP is opened on the endpoint port - since this is the only port needed to communicate with CAPC. Other tasks such as icmp checks, ntp sync, etc should be out of scope (since this is up to the implementation of the underlying OS and not CAPI specific) and handled as an additional feature instead of a mandatory requirement (as some users might not want these features or have extra rules open). IMO the user or any other application integrated with CAPC should open these FW rules since they aren't related to CAPC's working and the requirements can grow as the services running on the nodes increase

**Comment by chrisdoherty4** [comment]
> Discussed this on the issue. +1 with documentation

Do you have any feedback around where the documentation should live? We probably don't need to block this PR on docs though, we could just keep the issue open with a note until we figure out docs.

**Comment by weizhouapache** [comment]
@hrak 
please kindly ignore the test failure with capi-provider-cloudstack-presubmit-e2e-smoke-test

**Comment by hrak** [comment]
Ok, I am currently on holiday so I can't really address any issues. Back
next week.

Op wo 31 mei 2023 11:31 schreef Wei Zhou ***@***.***>:

**Comment by hrak** [comment]
Updated the PR with a note in the docs. Let me know if this will suffice.

File: docs/book/src/clustercloudstack/configuration.md [hunk 2]
```diff
@@ -61,7 +61,8 @@ cmk list zones listall=true | jq '.zone[] | {name, id}'
 
 The network must be declared as an environment variable `CLOUDSTACK_NETWORK_NAME` and is a mandatory parameter.
 As of now, only isolated and shared networks are supported.
-If the specified network does not exist, a new isolated network will be created.
+
+If the specified network does not exist, a new isolated network will be created. The newly created network will have a default egress firewall policy that allows all TCP, UDP and ICMP traffic from the cluster to the outside world.
```
davidjumani, line 65: Would be great to also mention that since CAPC might need access to the outside world, any additional rules outside of ACS might also need to be opened.

PR #230, opened 2023-04-03
Upgrade cluster-api dependency to v1.2.12

hrak:
*Issue #, if available:*

*Description of changes:*

This PR upgrades the cluster-api dependency to v1.2.12. Changelog [can be found here](https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.12).

After this we should probably look into upgrading to at least cluster-api 1.3.x since the 1.2.x release will be considered EOL now that 1.4.x is out.

*Testing performed:*

make build / make test / deployed on testing management cluster

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #242, opened 2023-05-08
e2e: fix "Failed to compile e2e"

weizhouapache:
*Issue #, if available:*

Got lots of errors when run "make run-e2e" like below
```
Failed to compile e2e:

/root/go/pkg/mod/sigs.k8s.io/cluster-api/test@v1.2.11/framework/cluster_helpers.go:231:12: cannot use input.Client (variable of type client.Client) as type Getter in struct literal:
	client.Client does not implement Getter (wrong type for Get method)
		have Get(ctx context.Context, key "k8s.io/apimachinery/pkg/types".NamespacedName, obj client.Object, opts ...client.GetOption) error
		want Get(ctx context.Context, key "k8s.io/apimachinery/pkg/types".NamespacedName, obj client.Object) error
```

The issue is because of one of the break changes between controller-runtime 0.12 and 0.13 see https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.13.0

*Description of changes:*

It can be fixed by downgrading sigs.k8s.io/controller-runtime from 0.13.1 to 0.12.3

*Testing performed:*


weizhouapache:
@rohityadavcloud @davidjumani @g-gaston @dims 
 
An alternative way to fix the e2e compilation error is, upgrade to cluster-api 1.4.2 (released last week)
However, it would be a major change.

hrak:
I would say merge this PR and #230 , then do a release. After that work on upgrading to cluster-api 1.3 or 1.4.

g-gaston:
@weizhouapache 
I'm cool merging this.

However, why does the main go.mod need to change? It seems like the breaking changes only affect the test module?

Also, what does everyone think about having a presubmit that builds the e2e tests? Just to make sure we don't break this again in the future.

hrak:
> However, why does the main go.mod need to change? It seems like the breaking changes only affect the test module?

I just checked CAPI v1.2.11 and v1.2.12, they depend on `sigs.k8s.io/controller-runtime v0.12.3` so this seems to have been inadvertently bumped. For as long as we depend on CAPI v1.2, `sigs.k8s.io/controller-runtime v0.12.3` seems to be correct.

> Also, what does everyone think about having a presubmit that builds the e2e tests? Just to make sure we don't break this again in the future.

I think this would be a good idea, but that would require a Cloudstack cluster somewhere to run the e2e tests.

weizhouapache:
> @weizhouapache I'm cool merging this.
> 
> However, why does the main go.mod need to change? It seems like the breaking changes only affect the test module?
> 
> Also, what does everyone think about having a presubmit that builds the e2e tests? Just to make sure we don't break this again in the future.



> > However, why does the main go.mod need to change? It seems like the breaking changes only affect the test module?
> 
> I just checked CAPI v1.2.11 and v1.2.12, they depend on `sigs.k8s.io/controller-runtime v0.12.3` so this seems to have been inadvertently bumped. For as long as we depend on CAPI v1.2, `sigs.k8s.io/controller-runtime v0.12.3` seems to be correct.
> 
thanks @hrak for the checking.
I was hesitated to create this PR because downgrade of a dependency is always not suggested in my opinion.
your check gives me more confidence that v0.12.3 might be a better version.

meanwhile, I am working on cluster-api 1.4.2, which works fine with the latest controller-runtime v0.14.5

> > Also, what does everyone think about having a presubmit that builds the e2e tests? Just to make sure we don't break this again in the future.
> 
> I think this would be a good idea, but that would require a Cloudstack cluster somewhere to run the e2e tests.

we can setup local environments in our lab, similar as we use jenkins to test all cloudstack PRs. 
cc @rohityadavcloud @davidjumani

g-gaston:
> I just checked CAPI v1.2.11 and v1.2.12, they depend on sigs.k8s.io/controller-runtime v0.12.3 so this seems to have been inadvertently bumped. For as long as we depend on CAPI v1.2, sigs.k8s.io/controller-runtime v0.12.3 seems to be correct.

That doesn't explain why the main module needs to change, does it? The fact that core capi v1.2.11 uses controller-runtime v0.12.3 doesn't mean that the core capi code is incompatible with newer versions of controller-runtime.

The capi test module is definitely incompatible with newer version of controller-runtime, but that's a separate module. And that module is only imported from the capc test/e2e module.

I'm just trying to understand the need for the change here :)

> I think this would be a good idea, but that would require a Cloudstack cluster somewhere to run the e2e tests.

> we can setup local environments in our lab, similar as we use jenkins to test all cloudstack PRs.

That would be awesome, but I was only suggesting doing a "dry-run" of the tests, at least as a first step. Like just checking that the code is valid at a compiler level. To make sure we don't break the code by bumping dependencies in the future.

weizhouapache:
> > I just checked CAPI v1.2.11 and v1.2.12, they depend on sigs.k8s.io/controller-runtime v0.12.3 so this seems to have been inadvertently bumped. For as long as we depend on CAPI v1.2, sigs.k8s.io/controller-runtime v0.12.3 seems to be correct.
> 
> That doesn't explain why the main module needs to change, does it? The fact that core capi v1.2.11 uses controller-runtime v0.12.3 doesn't mean that the core capi code is incompatible with newer versions of controller-runtime.
> 
@g-gaston 
currently we have not seen any other issues caused by controller-runtime v0.13.0. 
However, it is a blocker issue for e2e tests. there are two options to fix it
(1) downgrade controller-runtime to v0.12.3, or
(2) upgrade cluster-api (to v1.4.2) and other dependencies.

I've shared my thoughts in the decription of issue https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/246
this PR is not needed if we decide not to run e2e tests for 0.4.X releases.

> The capi test module is definitely incompatible with newer version of controller-runtime, but that's a separate module. And that module is only imported from the capc test/e2e module.
> 
> I'm just trying to understand the need for the change here :)
> 
> > I think this would be a good idea, but that would require a Cloudstack cluster somewhere to run the e2e tests.
> 
> > we can setup local environments in our lab, similar as we use jenkins to test all cloudstack PRs.
> 
> That would be awesome, but I was only suggesting doing a "dry-run" of the tests, at least as a first step. Like just checking that the code is valid at a compiler level. To make sure we don't break the code by bumping dependencies in the future.

@g-gaston 
if we just compile e2e, it is simple. it can be probably done by github actions.

I would suggest to run e2e tests on real cloudstack enviroments.

hrak:
> That doesn't explain why the main module needs to change, does it? The fact that core capi v1.2.11 uses controller-runtime v0.12.3 doesn't mean that the core capi code is incompatible with newer versions of controller-runtime.


Thats a fair point, the provider has been running fine in our env using the current 0.13.1 controller-runtime dep.

chrisdoherty4:
> (1) downgrade controller-runtime to v0.12.3

The last release [pegged controller-runtime at 0.11.1](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/8246eb304bcc89d7483984eae027713dbe58de7e/go.mod) so we aren't really downgrading anything. 

Lets make the dependencies make sense for CAPI 1.12.x and release as has been suggested, then focus on CAPI upgrades.

g-gaston:
It seems like controller-runtime was upgraded [here](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/189) to support k8s 1.24

Given the controller seem to have been working fine, if it's not required to fix the e2e tests, I vote to leave the main module untouched in order to reduce risk.

weizhouapache:
> It seems like controller-runtime was upgraded [here](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/189) to support k8s 1.24
> 
> Given the controller seem to have been working fine, if it's not required to fix the e2e tests, I vote to leave the main module untouched in order to reduce risk.

same as @chrisdoherty4 , I checked the go.mod of latest release (v0.4.8), the verison is v0.11.1
https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/v0.4.8/go.mod#L25

The v0.4.9-rc1/rc2/rc3/rc4 do use the newer version v0.13.1, but there is no official release v0.4.9 yet. 
https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/v0.4.9-rc4/go.mod#L25

Issue #244, opened 2023-05-08
Machine controller panics with logging bug


vignesh-goutham:
/kind bug

Cloudstack machine controller panics when trying to get generate affinity group name. Affinity group names are generated with the objects owner ref. If the object is owned by another entity than KCP, Etcdadm, or MachineSet, the AG group name generation fails. This causes the controller to panic with log.

Issue #240, opened 2023-05-05
Remove Affinity Group dependency on owner ref


vignesh-goutham:
/kind bug

Cloudstack machine spec accepts affinity group in two ways.
1. Managed way where we can just say `pro` or `anti` and CAPC will create an affinity group accordingly for each node group. Ex. for KCP, Etcd, and each worker node group.
2. Provide an already created list of affinity group IDs

In the managed way, CAPC gets the owner ref of CAPI machine and uses the owner name to identify and generate a unique name per node group. While this would work in most situations, owner ref of CAPI machines is not a solid or strict API spec to identify the node role or node group. For instance, ETCD machines can be owned by etcdamd controller, but when the controller chooses to delete old machines, it simply removes its ownership, at which point, CAPI cluster owns those old etcd machines. 

An intermediary API that is responsible to track node roles along with affinity or anti-affinity status should be established from which affinity group names should be generated. This API should derive the affinity details during KCP/Cluster creation time.

rohityadavcloud:
Fixed by https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/286 cc @vignesh-goutham pl review and test, thanks.

PR #245, opened 2023-05-08
Fix panic on machine controller when considering affinity

vignesh-goutham:
*Issue #, if available:*
#244

*Description of changes:*
Cloudstack machine controller supports creating and managing affinity groups. When the controller tries to generate an affinity group name it uses the owner ref of the CAPI machine object to identify node role and uses those owner ref to generate the AG group name. Controller only supports and expects CAPI machine to be owned by KCP, Etcdadm or MachineSet. If a CAPI machine is owned by CAPI cluster (some cases with etcd machine), this flow fails and results in an error. A bug with the logging causes panic. These changes fixes the panic for now.

Future scope on how to streamline the process to identify node roles without depending on CAPI machine's owner ref directly is discussed [here](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/240)

*Testing performed:*
Tested by removing ownerRef and monitoring if CAPC crashes.

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -128,7 +128,7 @@ func (r *CloudStackMachineReconciliationRunner) Reconcile() (retRes ctrl.Result,
 	)
 }
 
-// ConsiderAffinity sets machine affinity if needed. It also creates or gets an affinity group CRD if required and
+// ConsiderAffinity sets machine affinity if needed. It also creates or gets an affinity group resource if required and
 // checks it for readiness.
 func (r *CloudStackMachineReconciliationRunner) ConsiderAffinity() (ctrl.Result, error) {
 	if r.ReconciliationSubject.Spec.Affinity == infrav1.NoAffinity ||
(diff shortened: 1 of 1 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -138,7 +138,8 @@ func (r *CloudStackMachineReconciliationRunner) ConsiderAffinity() (ctrl.Result,
 
 	agName, err := utils.GenerateAffinityGroupName(*r.ReconciliationSubject, r.CAPIMachine)
 	if err != nil {
-		r.Log.Info("getting affinity group name", err)
+		r.Log.Info("getting affinity group name", err.Error())
```
g-gaston, line 141:
```suggestion
```

If the error is returned and bubbled up from the the `Reconcile` method, controller-runtime will print it by default. So I don't think we need this, so I vote to just remove.

Otherwise, we can just:

```suggestion
		r.Log.Error(err, "Failed getting affinity group name")
```

vignesh-goutham: Removed the log statement. The error gets bubbled up to runtime.

PR #250, opened 2023-05-11
Create unmanaged kubernetes cluster on CloudStack

vishesh92:
*Issue #, if available:*
*Description of changes:*
Creates an unmanaged kubernetes cluster on Cloudstack. This provides a nice view of CAPC managed kubernetes cluster in CloudStack UI.

Snapshots from ACS Dashboard:
Screenshot from 2023-08-22 17-24-21
Screenshot from 2023-08-22 17-24-30

Depends on: https://github.com/apache/cloudstack/pull/7515 https://github.com/apache/cloudstack-go/pull/59 (go.mod needs to be updated once this is merged)
*Testing performed:*
Tested locally by creating a kubernetes cluster using cluster api.


vishesh92:
Hi @chrisdoherty4
This creates an entry for a kubernetes cluster in CloudStack. This helps provide a good view of CAPC managed clusters on CloudStack and the VMs associated with the cluster in the CloudStack UI. This PR is dependent on `apache/cloudstack-go` PR merge and release. So, it will be some time before this PR can be merged.

chrisdoherty4:
@vishesh92 So populating this value would bind VMs to an [Apache CloudStack Cluster](http://docs.cloudstack.apache.org/en/latest/conceptsandterminology/concepts.html#about-clusters) ('cluster' is hyperloaded in this context)?

Am I right in saying we expect users to have pre-created the cluster?

config/.flag-test.mk
chrisdoherty4, line 1: Was this meant to be checked in?

vishesh92: My bad. let me remove it.

vishesh92:
`cluster` has a different meaning when it comes to cloudstack infrastructure. In context of this PR, `cluster` == `kubernetes cluster`.

We are doing two operation when the CAPC managed kubernetes cluster comes up:
1. Create an entry in CloudStack's kubernetes dashboard. This is just a simple insert into db and cloudstack doesn't do anything else on this resource.
2. When the machine deployments get created or deleted, we link or de-link the VM created by CAPC on cloudstack with kubernetes cluster (on CloudStack) entry created earlier.

In short, we are just adding some information about the kubernetes cluster which is managed by CAPC on CloudStack.

rohityadavcloud:
ping @vishesh92 is this up to date?

go.sum
```diff
@@ -448,6 +450,8 @@ github.com/rogpeppe/go-internal v1.6.1/go.mod h1:xXDCJY+GAPziupqXw64V24skbSoqbTE
 github.com/russross/blackfriday/v2 v2.0.1/go.mod h1:+Rmxgy9KzJVeS9/2gXHxylqXiyQDYRxCVz55jmeOWTM=
 github.com/ryanuber/columnize v0.0.0-20160712163229-9b3edd62028f/go.mod h1:sm1tb6uqfes/u+d4ooFouqFdy9/2g9QGwK3SQygK0Ts=
 github.com/sean-/seed v0.0.0-20170313163322-e2103e2c3529/go.mod h1:DxrIzT+xaE7yg65j358z/aeFdxmN0P9QXhEzd20vsDc=
+github.com/shapeblue/cloudstack-go/v2 v2.9.1-0.20230717062313-73e4efc8a510 h1:FPRBv784robz6sZSqDGfZDZMse31lj96i+enH02Xzds=
```
weizhouapache, line 453: @vishesh92 
the latest cloudstack-go verison is 2.15.0 (see https://github.com/apache/cloudstack-go)
2.13.1 was used in this project before.

can you update versions of shapeblue repo ?

vishesh92: We create a beta release of cloudstack-go and use that here instead

pkg/cloud/isolated_network.go
```diff
@@ -100,7 +100,8 @@ func (c *client) CreateIsolatedNetwork(fd *infrav1.CloudStackFailureDomain, isoN
 	}
 
 	// Do isolated network creation.
-	p := c.cs.Network.NewCreateNetworkParams(isoNet.Spec.Name, isoNet.Spec.Name, offeringID, fd.Spec.Zone.ID)
+	p := c.cs.Network.NewCreateNetworkParams(isoNet.Spec.Name, offeringID, fd.Spec.Zone.ID)
```
weizhouapache, line 103: this requires new cloudstack-go release as well

controllers/cloudstackcluster_controller.go
```diff
@@ -93,9 +94,28 @@ func (r *CloudStackClusterReconciliationRunner) Reconcile() (res ctrl.Result, re
...
+	if r.ShouldReturn(res, err) {
+		return res, err
+	}
+	err = r.CSUser.GetOrCreateCluster(r.CAPICluster, r.ReconciliationSubject, &r.FailureDomains.Items[0].Spec)
+	if err != nil {
+		if strings.Contains(err.Error(), "Kubernetes Service plugin is disabled") {
+			r.Log.Info("Kubernetes Service plugin is disabled on CloudStack. Skipping ExternalManaged kubernetes cluster creation")
+			return ctrl.Result{}, nil
+		}
+		// Not requeing the failure to support CloudStack v4.18 and before
```
weizhouapache, line 113: typo "requeing" ?

weizhouapache (approved):
overall code looks good to me.

davidjumani:
@vishesh92 This PR looks good, and you've added unit tests as well. Can you add a check in one of the e2e tests to ensure that the resources are created on ACS ?

davidjumani:
> `cluster` has a different meaning when it comes to cloudstack infrastructure. In context of this PR, `cluster` == `kubernetes cluster`.

I think it would be better to rename any reference to an ACS cluster as an unmanaged cluster (the interface, method names, etc) to avoid any confusion similar to how the API has CloudStackClusterID and not ClusterID

vishesh92:
@rohityadavcloud 
a) It shouldn't be an issue if the CKS plugin is not enabled or is on an older version.
b) I have introduced a cluster type field which is used to differentiate between the clusters.

vishesh92:
> It seems like this could be done by anything with access to the cloudstack API and the kube api server, right? We don't need special information that is only available inside the capc controller or to hook into particular events that are hidden in the reconcilers. This just needs to watch the `CloudstackCluster` resources and possibly the `CloudstackMachines`'s?

I want to make the integration a part of the CAPC controller since it has enough information about all the resources and the account and domain it needs to be part of. In future we might add even more information regarding the CAPC cluster in CKS as well and having it as part of the controller would help.

hrak:
> I might be missing the point of this feature but it feels a little bit like unnecessary complexity. It's not only just adding this feature, now we need to maintain it and test it and the benefits seem dubious to be.
> 
> Given that the CAPI API is our pane of glass to describe k8s clusters, adding a second one that works in parallel and not built on top of it appears like duplicated functionality.

I'm with you, this is adding a lot of CKS-specific logic to the controller, whose only task it should be to reconcile CAPI based clusters. I think there are bigger fish to fry (keeping up-to-date with CAPI and k8s versions, releasing more often etc.)

rohityadavcloud:
Hi @g-gaston @hrak I see your points and completely understand why would anybody want to make CAPC complex. But hear us out.

Apache CloudStack has always been a community-led, turnkey IaaS platform. The emphasis on turn-key, easy to use, setup and deploy, with a great UI and that it just works (out of the box). While CAPC can be used as a building-block, such as for EKS-A, we think this is also used directly as a first-class tool in itself with clusterctl.

Before CAPC, CloudStack's only supposed Kubernetes integration was in the form of CKS (CloudStack Kubernetes Service) wherein CloudStack orchestrates k8s clusters for its users. However, the CKS built clusters make some assumption and deploy specific CNIs to make this work out of the box - easy to deploy k8s cluster for most users. And CKS is available for all user account types, if the service providers, operator or enterprise admin enables and configures CKS in their environment.

With CAPC, there is more flexibility but this isn't available for all user types (currently, domain and root admins) and a group of contributors think it would be good idea if an end user can simply deploy a CAPC cluster and there is just a linked-resource created in the form of unmanaged CKS cluster (for example, to give you an idea head over to https://qa.cloudstack.cloud/simulator/#/kubernetes to see such a resource, this shoudn't be confused with a CloudStack Cluster of a zone).

To address these, we've created a general mechanism via https://github.com/apache/cloudstack/pull/7515 and series of fixes such the ability for normal user accounts to pick/use a public IP say in a shared network with https://github.com/apache/cloudstack/issues/7817. This could help fix https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/303 in future releases. The final lacking piece if a normal user account deploy a CAPC cluster, can we give him a control plane view to say there exists a k8s cluster in CloudStack but it's externally (CAPC) managed - that's where this feature comes in.

Now, you're both right - we don't need this PR to do that, there could be an external script and tool for this, however, it would make it complex for users and operators to set this up and the overall solution wouldn't be turnkey. Such an external tool/script would need to be aware of CloudStack and aware/access of the CAPC cluster and do that for all accounts would just introduce multiple competing control plane paths.

So, here's my proposal - @vishesh92 can introduce an new optional/parameter in the CAPC cluster generated yaml (or via a support cmd-line arg that generate the same) to enable/disable the feature. By default, this is kept enabled so any CAPC user deploying a k8s cluster see an unmanaged CKS cluster in the UI. And downstream tools and users who don't want this behaviour can turn it off via the option/setting in the generated yaml. This way you can control that the new code introduced in this PR doesn't get into the operational path in downstream use-cases, and upstream CAPC and ACS releases are turnkey.

Thoughts?

chrisdoherty4:
Hi @rohityadavcloud,

Thanks for sharing your thoughts and elaborating on intent. I do have reservations about this feature. Maintaining the scope outlined in the Cluster API documentation for infrastructure providers ensures we don't feature creep and have an easier time maintaining CAPC. Additional functionality, such as this PR, can be added through auxiliary components, as you highlighted, which would honor the single responsibility principle. Furthermore, considering the roadmap for CKS integration is somewhat unclear to the CAPC community, having a best effort integration doesn't feel like the right path forward.

> With CAPC, there is more flexibility but this isn't available for all user types ... it would be good idea if an end user can simply deploy a CAPC cluster and there is just a linked-resource created in the form of unmanaged CKS cluster

This sounds like 2 different asks:
1. "I want to deploy CAPC clusters using any user type"
2. "I want a cluster to appear in my CKS plugin UI"

On (1), I think we need to ask why said user types are required and broaden the scope we're considering. For example, perhaps a more flexible permissions model in Apache CloudStack would enable finer permission control so users can create whatever kind of user they wish?

On (2), I empathize with this desire, but I question what a user can do with an unmanaged CKS cluster (not to mention it isn't a CKS cluster) -  it's read-only if I'm not mistaken? Have we established why this is needed? In any case, it feels like it falls outside the scope of CAPC.

> however, it would make it complex for users and operators to set this up and the overall solution wouldn't be turnkey

I think that's dependent on how you create a solution. Cluster API in itself isn't turnkey: you have to deploy Cluster API, then you can create your clusters. There's nothing preventing building a solution that wraps deployment of all the necessary components, including a component that embodies the integration in this PR, which would be truly turnkey; though I don't think that's a concern for CAPC.

> Such an external tool/script would need to be aware of CloudStack and aware/access of the CAPC cluster and do that for all accounts would just introduce multiple competing control plane paths.

That seems like an appropriate set of dependencies for this kind of component, though I don't understand the "do it for all accounts" or "competing control plane paths" parts; perhaps you could elaborate?

I think its worth pointing out that all existing components built for Kubernetes have this kind of dependency set. Monitoring tools like Jaeger and Prometheus come to mind.

chrisdoherty4:
Hi @vishesh92. It sounds like you may have tried gathering feedback prior to completing the upstream CloudStack work and the work in this PR. I'm sorry I missed that and have provided feedback late in the day including leaving this PR untouched for a prolonged period. We should be aiming to avoid this again, lets chat in the Kubernetes CAPC Slack on how to best address the communication inefficiencies.

PR #254, opened 2023-05-16
Fix userdata compression

chrisdoherty4:
# Summary

When the UncompressedUserData field is nil on CloudStackMachine instances CAPC is interpreting it as true.

This reworks the compression decision logic to give a nil UncompressedUserData the same semantics as false.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,7 +263,8 @@ config/.flag-test.mk: $(CONTROLLER_GEN) $(MANIFEST_GEN_INPUTS_TEST)
 	@touch config/.flag-test.mk
 
 .PHONY: test
-test: generate-deepcopy-test generate-manifest-test generate-mocks lint $(GINKGO_V2) $(KUBECTL) $(API_SERVER) $(ETCD) ## Run tests. At the moment this is only unit tests.
+test: ## Run tests.
+test: generate-deepcopy-test generate-manifest-test generate-mocks lint $(GINKGO_V2) $(KUBECTL) $(API_SERVER) $(ETCD)
 	@./hack/testing_ginkgo_recover_statements.sh --add # Add ginkgo.GinkgoRecover() statements to controllers.
 	@# The following is a slightly funky way to make sure the ginkgo statements are removed regardless the test results.
 	@$(GINKGO_V2) --label-filter="!integ" --cover -coverprofile cover.out --covermode=atomic -v ./api/... ./controllers/... ./pkg/...; EXIT_STATUS=$$?;\
--- a/pkg/cloud/helpers.go
+++ b/pkg/cloud/helpers.go
@@ -18,7 +18,7 @@ package cloud
 
 import (
 	"bytes"
-	"compress/gzip"
+	cgzip "compress/gzip"
 )
 
 type set func(string)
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -263,7 +263,8 @@ config/.flag-test.mk: $(CONTROLLER_GEN) $(MANIFEST_GEN_INPUTS_TEST)
 	@touch config/.flag-test.mk
 
 .PHONY: test
-test: generate-deepcopy-test generate-manifest-test generate-mocks lint $(GINKGO_V2) $(KUBECTL) $(API_SERVER) $(ETCD) ## Run tests. At the moment this is only unit tests.
```
chrisdoherty4, line 266: Removed the comment about '[running] only unit tests' because the recipe runs a host of integration tests that need to be re-written into proper unit tests.

pkg/cloud/instance.go
```diff
@@ -346,15 +372,3 @@ func (c *client) listVMInstanceDatadiskVolumeIDs(instanceID string) ([]string, e
 	return ret, nil
 
 }
-
-// handleUserData optionally compresses and then base64 encodes userdata
-func handleUserData(userData string, uncompressed *bool) (string, error) {
```
chrisdoherty4, line 351: This function felt unnecessary, I've moved the logic to the consumer sight.

pkg/cloud/instance.go
```diff
@@ -268,21 +270,19 @@ func (c *client) GetOrCreateVMInstance(
...
-		listVirtualMachineParams := c.cs.VirtualMachine.NewListVirtualMachinesParams()
-		listVirtualMachineParams.SetTemplateid(templateID)
-		listVirtualMachineParams.SetZoneid(fd.Spec.Zone.ID)
-		listVirtualMachineParams.SetNetworkid(fd.Spec.Zone.Network.ID)
-		listVirtualMachineParams.SetName(csMachine.Name)
-		listVirtualMachinesResponse, err2 := c.cs.VirtualMachine.ListVirtualMachines(listVirtualMachineParams)
-		if err2 != nil || listVirtualMachinesResponse.Count <= 0 {
-			c.customMetrics.EvaluateErrorAndIncrementAcsReconciliationErrorCounter(err2)
-			return err
+		// CloudStack may have created the VM even though it reported an error. We attempt to
```
chrisdoherty4, line 273: I refactored this because the `GetOrCreateVMInstance()` method had a high cyclomatic complexity.

pkg/cloud/helpers.go
```diff
@@ -43,13 +43,13 @@ func setIntIfPositive(num int64, setFn setInt) {
 	}
 }
 
-func CompressString(str string) (string, error) {
-	buf := &bytes.Buffer{}
-	gzipWriter := gzip.NewWriter(buf)
-	if _, err := gzipWriter.Write([]byte(str)); err != nil {
+func zip(str string) (string, error) {
```
g-gaston, line 46: nit: should this be?
```suggestion
func gzip(str string) (string, error) {
```

pkg/cloud/helpers_test.go
```diff
@@ -17,34 +17,14 @@ limitations under the License.
...
 	"github.com/golang/mock/gomock"
-	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
-	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
 )
 
-var _ = Describe("Helpers", func() {
-
-	It("should compress and encode string", func() {
```
g-gaston, line 35: is the gzip thing being tested through some other exported method?

g-gaston (commented):
lgtm

pkg/cloud/helpers_test.go
chrisdoherty4: Yes, the tests have been modified to check across multiple tests. Ideally there would be a unit test just for this but the way the code is structured makes that challenging.

g-gaston (commented):
/lgtm

PR #258, opened 2023-05-25
e2e: upgrade k8s versions in test/e2e/config/cloudstack.yaml

weizhouapache:
*Issue #, if available:*

This is used to fix the failure with e2e test test/e2e/kubernetes_version_upgrade_test.go

The KubeadmControlPlane "kubernetes-version-upgrade-cko019-control-plane" is invalid: spec.version: Forbidden: cannot upgrade to a Kubernetes/kubeadm version which is using the old default registry. Please use a newer Kubernetes patch release which is using the new default registry (>= v1.22.17, >= v1.23.15, >= v1.24.9)

*Description of changes:*

Upgrade both old version and new version of k8s upgrade.

*Testing performed:*


weizhouapache:
The E2E failures are not related to this PR. This PR is ready for review 
cc @rohityadavcloud @chrisdoherty4 

The templates for k8s v1.24.14 have been built, but not published yet.
We have also built images for v1.25.10, v1.26.5, v1.27.2. We will test and publish the images.

PR #264, opened 2023-06-01
Makefile cleanup for CRD, webhooks, rbac, conversion, etc generation

vignesh-goutham:
*Description of changes:*
Few bugs in makefile caused conversions to not be generated on API changes including some that are already merged. This change fixes those bugs and cleans up makefile to standardize webhook, crd, rbac and mocks generation.

Also has changes to include kubebuilder tags to prevent any new conversion changes and keep API as already checked-in. We have to make sure future API versions support conversions back and forth between Hub and Spoke and are lossless as well, as we have many users running `v1beta2` already.

*Testing performed:*
Ran `make generate` to verify crds, webhooks, etc get generated properly.


api/v1beta2/cloudstackmachine_types.go
```diff
@@ -139,10 +140,12 @@ type CloudStackMachineStatus struct {
 
 	// Status indicates the status of the provider resource.
 	// +optional
+	// +k8s:conversion-gen=false
 	Status *string `json:"status,omitempty"`
```
g-gaston, line 144: curious why can't we convert the status?, is this a new field?

config/webhook/manifests.yaml
```diff
@@ -92,23 +92,3 @@ webhooks:
     resources:
     - cloudstackmachines
   sideEffects: None
-- admissionReviewVersions:
```
g-gaston, line 95: was this unnecessary?

g-gaston (commented):
changes in make file look good, just curious about the changes in conversion

config/webhook/manifests.yaml
weizhouapache: seems so. this file is changed every time when run `make` command.

api/v1beta2/cloudstackmachine_types.go
weizhouapache: yes, there are 3 new fields in v1beta2

chrisdoherty4:
Looks like this isn't quite ready. Feel free to re-request when ready.

/uncc chrisdoherty4

rohityadavcloud:
Hi @vignesh-goutham is this still valid/required, can you fix conflicts?

vignesh-goutham:
Will re-open the PR once ready.

Issue #266, opened 2023-06-05
"post-cluster-api-provider-cloudstack-push-images" job failed for 201 days


weizhouapache:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

https://github.com/kubernetes-sigs/cluster-api/issues/8784

The log of latest run
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-cluster-api-provider-cloudstack-push-images/1664572298696331264

**What did you expect to happen:**

succeed.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

It is because the docker container does not have gcc installed.

**Environment:**

- Cluster-api-provider-cloudstack version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

g-gaston:
@weizhouapache any clue what part of the code is using cgo?

PR #268, opened 2023-06-08
Disable CGO when running generate deep copy

g-gaston:
*Issue #, if available:* 
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/266

*Description of changes:*
This should fix the job that periodically builds images, since the container where the releas staging target is run doesn't have gcc installed.

*Testing performed:*
Ran `make generate-deepcopy` inside `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20220609-2e4c91eb7e` and verified it works.

Inspiration from https://github.com/kubernetes-sigs/cluster-api/issues/6230


weizhouapache:
verified ok.
`make docker-build` works as well

/lgtm

PR #312, opened 2023-09-06
Propagate error from erroneous VM deployment

chrisdoherty4:
When creating virtual machines the CloudStack API call may fail but a VM is still created. In this case, we populate VM data on the CloudStackMachine object so the end user can manually clean up. However, we don't return the underlying error that results in the error being swallowed.


rohityadavcloud:
@chrisdoherty4 is this ready for testing?

/run-e2e

PR #313, opened 2023-09-22
Stop swallowing VM deployment error

chrisdoherty4:
When a VM fails to deploy we swallow the original error resulting in no controller logs entries and no events being generated. Returning the error ensures both occur and provides operator with an indication of the problem.


g-gaston (commented):
/lgtm

Issue #321, opened 2023-10-31
Domain resolution assumes `ROOT` as parent domain


vignesh-goutham:
/kind bug

**What steps did you take and what happened:**
Domain resolution is done at multiple places in each reconcile loop to resolve the domain parameters with the any of the input domain identifying details. These domain identifying details can be domain path, domain id, or domain name. 

When domain name is used to resolve, there is an assumption in the controller code that assumes the parent domain is `ROOT` and hence the path of the domain starts with `ROOT/`. This assumption is not true and must be fixed. Further, CAPC controllers should be verified that they do not pass in domain name as an alternative for domain path while resolving domains.

Code with assumption is [here](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/main/pkg/cloud/user_credentials.go#L79-L83)

rohityadavcloud:
Is this still an issue @vignesh-goutham ?

PR #322, opened 2023-10-31
Use Domain ID instead of path while resolving User

vignesh-goutham:
*Issue #, if available:* Does not fix but alleviates #321 in few code paths.

*Description of changes:*
The ListUsers API only returns a domain ID and a domain name pertaining to the user provided as input. Using domain name in place of domain path, leads to assuming the parent domain here is `ROOT`. Instead this change, sets the domain id which is more deterministic and leads to clear identification and resolution of domain.


g-gaston:
lgtm
/ok-to-test

Is this difficult to test through unit tests?

chrisdoherty4 (approved):

PR #338, opened 2024-01-17
Update ttlcache library to v3, disable TTL extension on hit

### LINKED ISSUE #336: Account/domain limits stay cached forever
/kind bug

**What steps did you take and what happened:**

The CloudStack client in `pkg/cloud` is using https://github.com/jellydator/ttlcache/ to cache certain data fetched from the CS API. When the client connects for the first time, it fetches account/domain limits and stores them in the TTL cache.

The ttlcache package is configured with [`SkipTtlExtensionOnHit(false)`](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/207f0a2310f1b6df181d5a9ef8162ad1c1005ec3/pkg/cloud/client.go#L257), meaning that as long as the cached account/domain limits are being hit by the client, their TTL never expires.

This is causing problems when f.e. a domain is out of a certain resource (CPU cores/memory f.e.) on initial hit & cache of the account/domain limits, and the operator frees some resources in the domain. CAPC will not pick up the newly freed/available resources, causing VMs to never start up.

This could easily be solved by setting `SkipTtlExtensionOnHit` to true.

**What did you expect to happen:**

CAPC should cache account/domain limits only for a certain time, not extend the TTL upon cache hit.

**Anything else you would like to add:**

I will address this with a PR shortly.

**Environment:**

- Cluster-api-provider-cloudstack version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

hrak:
*Issue #, if available:*

Fixes #336

*Description of changes:*

This PR gives the ttlcache library a major update and disables TTL extension on cache item hit.

*Testing performed:*

make test

### DIFF
--- a/pkg/cloud/client.go
+++ b/pkg/cloud/client.go
@@ -29,8 +29,8 @@ import (
 	"gopkg.in/yaml.v3"
 	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/metrics"
 
-	"github.com/ReneKroon/ttlcache"
 	"github.com/apache/cloudstack-go/v2/cloudstack"
+	ttlcache "github.com/jellydator/ttlcache/v3"
 	"github.com/pkg/errors"
 )
 
@@ -71,7 +71,7 @@ type SecretConfig struct {
 	StringData Config            `yaml:"stringData"`
 }
 
-var clientCache *ttlcache.Cache
+var clientCache *ttlcache.Cache[string, *client]
 var cacheMutex sync.Mutex
 
 var NewAsyncClient = cloudstack.NewAsyncClient
@@ -159,7 +159,8 @@ func NewClientFromConf(conf Config, clientConfig *corev1.ConfigMap) (Client, err
 	}
 
 	clientCacheKey := generateClientCacheKey(conf)
-	if client, exists := clientCache.Get(clientCacheKey); exists {
+	if item := clientCache.Get(clientCacheKey); item != nil {
+		var client interface{} = item.Value()
 		return client.(Client), nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cloud/client.go
```diff
@@ -29,8 +29,8 @@ import (
 	"gopkg.in/yaml.v3"
 	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/metrics"
 
-	"github.com/ReneKroon/ttlcache"
 	"github.com/apache/cloudstack-go/v2/cloudstack"
+	ttlcache "github.com/jellydator/ttlcache/v3"
```
g-gaston, line 33: nit: is the alias needed?

pkg/cloud/client.go
```diff
@@ -159,7 +159,8 @@ func NewClientFromConf(conf Config, clientConfig *corev1.ConfigMap) (Client, err
 	}
 
 	clientCacheKey := generateClientCacheKey(conf)
-	if client, exists := clientCache.Get(clientCacheKey); exists {
+	if item := clientCache.Get(clientCacheKey); item != nil {
+		var client interface{} = item.Value()
 		return client.(Client), nil
```
g-gaston, line 164: this is more for my own understanding than a request for change
But, why is casting required if`ttlcache.Cache` is parametrized with the types for key and value?

g-gaston (commented):
thanks for the fix!

pkg/cloud/client.go
vishesh92: +1. We don't need the alias here.

pkg/cloud/client.go
vishesh92, line 164: Casting was required with previous version of ttlcache. We should be able to directly return `item.Value()` here.
```suggestion
		return item.Value(), nil
```

vishesh92:
@hrak some minor code changes required.

pkg/cloud/client.go
vishesh92, line 33:
```suggestion
	"github.com/jellydator/ttlcache/v3"
```

vishesh92 (approved):

Issue #310, opened 2023-09-04
Reconcile a CAPC cluster to a CKS cluster


rohityadavcloud:
/kind feature

**Describe the solution you'd like**

CAPC reconciles various resources in CloudStack, we want to see CAPC leverage and reconcile notion of a CAPC k8s cluster to a CKS (unmanaged) cluster.

The CKS cluster resource in CloudStack captures notion of a k8s cluster in CloudStack, and recently a feature improvement made it possible to have notion of unmanaged k8s cluster with https://github.com/apache/cloudstack/pull/7515 with documentation at https://github.com/apache/cloudstack-documentation/pull/315 and API support via go-sdk change https://github.com/apache/cloudstack-go/pull/59. A PR proposed https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/250 which couldn't satisfy review from other stakeholders. With this issue, alternative options and approaches can be discussed or the previous PR reimagined.

As a CAPC user I would want the following:

- Ability to see my CAPC cluster show up as unmanaged CKS cluster in CloudStack (API and UI)
- Be able to see usage, of my CAPC cluster via CKS cluster (cpu, memory etc.)
- As MachineSets are added/removed, the related instances can be added removed in CKS 'unmanaged' cluster
- In large environment, esp a typical CI/CD use-cases, if as a user I lose/forget my CAPC admin cluster I should be able to manually delete the CAPC cluster resources (namely, VMs and networks)
- I should be able to do this as a user account type (note: a CloudStack role has a list of allowed/disallowed APIs, however, an account type control ability of an account type to do certain things such as acquire public IP of choice in a given shared/isolated/vpc network etc. needed typically for CAPC cluster)
- Prefer not to use another tool, than CAPC for reconciliation of a CAPC cluster to a CloudStack Kubernetes Service (CKS) cluster

Brief design doc: (to be expanded by @vishesh92)

chrisdoherty4:
Lots of relevant conversation has already taken place on https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/250.

g-gaston:
Moving to v0.5.0 since this is a new feature (and requires some more discussion if I recall correctly) and we are only prioritizing vital bugfixes for v0.4.9

rohityadavcloud:
[...]

PR #350, opened 2024-03-22
Reconcile CAPC Cluster as CloudStack CKS Cluster (Default: Opt Out)

rohityadavcloud:
*Issue #, if available:*

https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/issues/310

*Description of changes:*

The notion of a CAPC cluster represents a k8s cluster created and managed by CAPC. While a similar construct exists in CloudStack in form of CKS (CloudStack K8S Service) cluster. This PR aims to address #310 by introducing logic that can reconcile a CAPC cluster to a CKS cluster, while having this feature disabled by default (i.e. opt out by default). This will be useful for users (and integrations) who want to be aware about a potentially externally managed K8S cluster in CloudStack via the reconciled CKS cluster.

ACS 4.19 and later support such reconciliation with introduction of feature support PRs in upstream CloudStack releases:
https://github.com/apache/cloudstack-go/pull/59
https://github.com/apache/cloudstack/pull/7515
https://github.com/apache/cloudstack-documentation/pull/315

Previously attempted as https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/250

*Testing performed:*

To be updated


rohityadavcloud:
Currently - in progress, draft - not fit or ready for review yet.

vishesh92:
@g-gaston Please review this PR. I am targeting this PR for v0.5.0.

weizhouapache (approved):
overall lgtm

rohityadavcloud:
@g-gaston this looks like is ready for merging, the tests have passed. We'll keep this open for another week if you're able to help review this, thanks.

g-gaston:
> @g-gaston this looks like is ready for merging, the tests have passed. We'll keep this open for another week if you're able to help review this, thanks.

Apologies for the delay! Yes i will get to this next week

g-gaston:
Sorry again for the delay!

The API and implementation logic makes sense, however I would love to talk about the controller architecture. I would like to propose to implement this in a separate controller/s instead of adding extra if/else condition to the existing controllers.

IMO this has several advantages:
* It decouples this new feature from the critical path for the normal cluster reconciliation path. That's the main responsibility of this provider and the one we should prioritize among all others. In practice, it reduces the chances of introducing issues/bugs in that path and it avoids extra latency.
* It keeps a cleaner separation of concerns, so the main Cluster/Machine controllers only care about the CAPI provider responsibilities. IMO this makes the code cleaner and easier to follow.
* It makes this functionality modular: it opens the possibility of running this functionality outside of the CAPC controller manager, as something that builds on top of CAPC and not necessarily part of CAPC. It also allows cluster admins to enable/disable this functionality at the controller manager level, which for a new feature is quite useful.

Opinions?

vishesh92:
> Sorry again for the delay!
> 
> The API and implementation logic makes sense, however I would love to talk about the controller architecture. I would like to propose to implement this in a separate controller/s instead of adding extra if/else condition to the existing controllers.
> 
> IMO this has several advantages:
> 
>     * It decouples this new feature from the critical path for the normal cluster reconciliation path. That's the main responsibility of this provider and the one we should prioritize among all others. In practice, it reduces the chances of introducing issues/bugs in that path and it avoids extra latency.
> 
>     * It keeps a cleaner separation of concerns, so the main Cluster/Machine controllers only care about the CAPI provider responsibilities. IMO this makes the code cleaner and easier to follow.
> 
>     * It makes this functionality modular: it opens the possibility of running this functionality outside of the CAPC controller manager, as something that builds on top of CAPC and not necessarily part of CAPC. It also allows cluster admins to enable/disable this functionality at the controller manager level, which for a new feature is quite useful.
> 
> 
> Opinions?

@g-gaston Makes sense. I have done the changes you requested. Please review again.

api/v1beta1/conversion.go
```diff
@@ -51,7 +51,7 @@ func Convert_v1beta1_CloudStackCluster_To_v1beta3_CloudStackCluster(in *CloudSta
 //nolint:golint,revive,stylecheck
 func Convert_v1beta3_CloudStackCluster_To_v1beta1_CloudStackCluster(in *v1beta3.CloudStackCluster, out *CloudStackCluster, scope conv.Scope) error {
 	if len(in.Spec.FailureDomains) < 1 {
-		return fmt.Errorf("v1beta3 to v1beta1 conversion not supported when < 1 failure domain is provided. Input CloudStackCluster spec %s", in.Spec)
+		return fmt.Errorf("v1beta3 to v1beta1 conversion not supported when < 1 failure domain is provided. Input CloudStackCluster spec %s", fmt.Sprintf("%v", in.Spec))
```
g-gaston, line 54: why do we need another sprintf? can't we use Errorf with a %v?

api/v1beta3/cloudstackcluster_types.go
```diff
@@ -34,6 +34,10 @@ type CloudStackClusterSpec struct {
 
 	// The kubernetes control plane endpoint.
 	ControlPlaneEndpoint clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
+
+	// SyncWithACS determines if an externalManaged CKS cluster should be created on ACS.
+	// +optional
+	SyncWithACS bool `json:"syncWithACS,omitempty"`
```
g-gaston, line 40: nit: I could be wrong bu I thought the recommended way was to use pointers for optional fields? I think that's what I remember from the k8s api conventions

controllers/cks_cluster_controller.go
```diff
@@ -0,0 +1,124 @@
...
+func (reconciler *CksClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (retRes ctrl.Result, retErr error) {
+	r := NewCksClusterReconciliationRunner()
+	r.UsingBaseReconciler(reconciler.ReconcilerBase).ForRequest(req).WithRequestCtx(ctx)
+	r.WithAdditionalCommonStages(r.GetFailureDomains(r.FailureDomains))
+	return r.RunBaseReconciliationStages()
+}
+
+// Reconcile actually reconciles the CloudStackCluster.
+func (r *CksClusterReconciliationRunner) Reconcile() (res ctrl.Result, reterr error) {
+	if r.CSCluster.Spec.SyncWithACS {
```
g-gaston, line 73: nit: in go is more common to express this as a guard condition
```suggestion
	if !r.CSCluster.Spec.SyncWithACS {
	  	return ctrl.Result{}, nil
	}
```

It has the added benefit of reducing the indentation of the main business logic and imo makes reading it easier since you don't have to hold the if condition in your head

controllers/cks_cluster_controller.go
```diff
@@ -0,0 +1,124 @@
...
+		// Prevent premature deletion.
+		controllerutil.AddFinalizer(r.ReconciliationSubject, CksClusterFinalizer)
+
+		if len(r.FailureDomains.Items) == 0 {
+			return r.RequeueWithMessage("No failure domains found")
+		}
+
+		res, err := r.AsFailureDomainUser(&r.FailureDomains.Items[0].Spec)()
```
g-gaston, line 81: can we guarantee all clustes have at least one failure domain? else this could panic

controllers/cks_cluster_controller.go
```diff
@@ -0,0 +1,124 @@
...
+		if r.ShouldReturn(res, err) {
+			return res, err
+		}
+
+		r.Log.Info("Creating entry with CKS")
+		err = r.CSUser.GetOrCreateCksCluster(r.CAPICluster, r.ReconciliationSubject, &r.FailureDomains.Items[0].Spec)
+		if err != nil {
+			if strings.Contains(err.Error(), "Kubernetes Service plugin is disabled") {
+				r.Log.Info("Kubernetes Service plugin is disabled on CloudStack. Skipping ExternalManaged kubernetes cluster creation")
+				return ctrl.Result{}, nil
```
g-gaston, line 91: should we just return an error? if the plugin is not enabled in CS but the cluster is configured to use CKS cluster, this is a configuration error, right? or at least log it as an error

controllers/cks_cluster_controller.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+			r.Log.Info(fmt.Sprintf("Failed creating ExternalManaged kubernetes cluster on CloudStack. Error: %s", err.Error()))
```
g-gaston, line 93: logr favors using key values instead of string interpolation. also wouldn't controller runtime log this error if we are returning an error?
```suggestion
			r.Log.Info("Failed creating ExternalManaged kubernetes cluster on CloudStack", "error", err.Error()))
```
I think Log also has log.Error

controllers/cks_cluster_controller_test.go
```diff
@@ -0,0 +1,69 @@
...
+			}).MinTimes(1).Return(nil)
+			mockCloudClient.EXPECT().ResolveZone(gomock.Any()).AnyTimes()
+			mockCloudClient.EXPECT().ResolveNetworkForZone(gomock.Any()).AnyTimes().Do(
+				func(arg1 interface{}) {
+					arg1.(*infrav1.CloudStackZoneSpec).Network.ID = "SomeID"
+					arg1.(*infrav1.CloudStackZoneSpec).Network.Type = cloud.NetworkTypeShared
+				}).MinTimes(1)
+		})
+
+		It("Should create a cluster in CKS.", func() {
```
g-gaston, line 50: test for delete?

controllers/cks_machine_controller.go
```diff
@@ -0,0 +1,107 @@
...
+	r.UsingBaseReconciler(reconciler.ReconcilerBase).ForRequest(req).WithRequestCtx(ctx)
+	r.WithAdditionalCommonStages(
+		r.GetFailureDomainByName(func() string { return r.ReconciliationSubject.Spec.FailureDomainName }, r.FailureDomain),
+		r.AsFailureDomainUser(&r.FailureDomain.Spec))
+	return r.RunBaseReconciliationStages()
+}
+
+// Reconcile actually reconciles the CloudStackCluster.
+func (r *CksMachineReconciliationRunner) Reconcile() (res ctrl.Result, reterr error) {
+	if r.CSCluster.Spec.SyncWithACS {
```
g-gaston, line 69: same here about the guard

controllers/cks_machine_controller.go
```diff
@@ -0,0 +1,107 @@
...
+	r := NewCksMachineReconciliationRunner()
+	r.UsingBaseReconciler(reconciler.ReconcilerBase).ForRequest(req).WithRequestCtx(ctx)
+	r.WithAdditionalCommonStages(
+		r.GetFailureDomainByName(func() string { return r.ReconciliationSubject.Spec.FailureDomainName }, r.FailureDomain),
+		r.AsFailureDomainUser(&r.FailureDomain.Spec))
+	return r.RunBaseReconciliationStages()
+}
+
+// Reconcile actually reconciles the CloudStackCluster.
+func (r *CksMachineReconciliationRunner) Reconcile() (res ctrl.Result, reterr error) {
```
g-gaston, line 68: do we not want a finalizer on the CS machines as well?

controllers/cloudstackmachine_controller_test.go
```diff
@@ -54,6 +54,11 @@ var _ = Describe("CloudStackMachineReconciler", func() {
 			// Setup a failure domain for the machine reconciler to find.
 			Ω(k8sClient.Create(ctx, dummies.CSFailureDomain1)).Should(Succeed())
 			setClusterReady(k8sClient)
+
+			mockCloudClient.EXPECT().GetOrCreateCksCluster(gomock.Any(), gomock.Any(), gomock.Any()).Do(
```
g-gaston, line 58: do we need this?

main.go
```diff
@@ -267,4 +267,12 @@ func setupReconcilers(ctx context.Context, base utils.ReconcilerBase, opts manag
...
 		os.Exit(1)
 	}
+	if err := (&controllers.CksClusterReconciler{ReconcilerBase: base}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "CKSClusterController")
+		os.Exit(1)
+	}
+	if err := (&controllers.CksMachineReconciler{ReconcilerBase: base}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "CKSMachineController")
+		os.Exit(1)
+	}
```
g-gaston, line 277: could we make this optional? either a flag or some other type of config that enables these controllers
And IMO it should be opt-in for now, so users who are not interested on this don't run these controllers by default

g-gaston (commented):
Thanks so much for the refactor, looking great!

controllers/cks_cluster_controller.go
vishesh92: We can't. That is why I have added a check in Line 77.

controllers/cks_machine_controller.go
vishesh92: We can have a finalizer. But if the VM gets destroyed/expunged from ACS, it automatically gets removed from the CKS cluster as well.

api/v1beta3/cloudstackcluster_types.go
vishesh92: As per my understanding we don't need to do that here. Pointer is used if we want to distinguish between explicit `false` and not specified.

Ref.: https://book.kubebuilder.io/cronjob-tutorial/api-design.html - search for `pointer` on the page.

controllers/cloudstackmachine_controller_test.go
vishesh92: We need it for the `CksClusterReconciler`.

vishesh92:
@g-gaston I have addressed the comments.

api/v1beta3/cloudstackcluster_types.go
g-gaston: I was talking about this: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

Again not required, but k8s does seem to recommend using pointer if the type doesnt have a built-in nil.

On the topic of needing to distinguish unset vs false:

> Using a pointer allows distinguishing unset from the zero value for that type. There are some cases where, in principle, a pointer is not needed for an optional field since the zero value is forbidden, and thus implies unset. There are examples of this in the codebase. However:
>
>*    it can be difficult for implementors to anticipate all cases where an empty value might need to be distinguished from a zero value
>*   structs are not omitted from encoder output even where omitempty is specified, which is messy;
>*   having a pointer consistently imply optional is clearer for users of the Go language client, and any other clients that use corresponding types

controllers/cks_cluster_controller.go
```diff
@@ -0,0 +1,120 @@
...
+// Reconcile actually reconciles the CloudStackCluster.
+func (r *CksClusterReconciliationRunner) Reconcile() (res ctrl.Result, reterr error) {
+	if !r.CSCluster.Spec.SyncWithACS {
+		return ctrl.Result{}, nil
+	}
+	// Prevent premature deletion.
+	controllerutil.AddFinalizer(r.ReconciliationSubject, CksClusterFinalizer)
+
+	if len(r.FailureDomains.Items) == 0 {
+		return r.RequeueWithMessage("No failure domains found")
```
g-gaston, line 80: should this be an error? or maybe a no-op?
Why do we reconcile? it seems like the user would need to make changes for this to start working, right? Maybe I'm missing something

controllers/cks_machine_controller.go
g-gaston: Ahh got you. Then do we even need the reconcile delete?

controllers/cloudstackmachine_controller_test.go
g-gaston: Could you help me understand this? I'm a bit lost. Why do we need to add a mock expectation on a test for one controller when the mock is being called in a different controller?

controllers/cks_machine_controller.go
vishesh92: I added it in case the functionality upstream changes. So, this is good to have.

controllers/cloudstackmachine_controller_test.go
vishesh92: My bad. We don't need it. I mistook your comment for `cks_machine_controller_test.go`. removing this.

controllers/cks_cluster_controller.go
vishesh92: Makes sense. Let me do no-op here to match the behavior with cloudstack controller.

controllers/cks_machine_controller.go
g-gaston: ok sounds good. but then i think we should have a finalizer
Not trying to be too purist, but if we are adding code to ensure the system works, even if not fully necessary right now, we should add it in a way that actually ensures the system works. If upstream changes and we don't have a finalizer, the cs vm can be deleted before we get a chance to execute this.

vishesh92: @g-gaston I have added the finalizer.

api/v1beta3/cloudstackcluster_types.go
g-gaston: @vishesh92 are you planning on changing this? or what's the decision here?
Otherwise the PR looks good to go

vishesh92: I was thinking of keeping it as bool only.

vishesh92: @g-gaston I have updated the PR to make it a pointer to a bool. Please check now.

g-gaston (commented):
/lgtm

Thanks a lot!

vishesh92:
> /lgtm
> 
> Thanks a lot!
> 
> I see you are trying to run tests. Let me know when you are done and ready to merge and I'll approve

Thanks @g-gaston for review. @kiranchavala is currently testing this PR. Once he gives a go ahead, we can merge this PR. Meanwhile, I will fix the conflicts in go.mod.

rohityadavcloud:
@vishesh92 while I opened the PR, it’s largely your work. Pl feel free to resubmit a new PR or lgtm this when this is ready after QA. cc @kiranchavala

kiranchavala (approved):
LGTM,

Tested manually and 


A externally managed kubernetes  cluster got created in cloudstack when a capc deployment was initiated 

Screenshot 2024-06-24 at 4 07 04 PM

Screenshot 2024-06-24 at 4 07 33 PM

Issue #352, opened 2024-03-22
Failure domain re-assignment on Cloudstack machine deploy failures


vignesh-goutham:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
CAPC chooses a [random failure domain](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/d597e8056e82ac6418171e66fa2532b2daaef987/controllers/cloudstackmachine_controller.go#L185) to deploy worker machines. When this VM deploy fails, irrespective of the type of error, CAPC will keep re-attempting to deploy a VM until CAPI replaces the owner machine. 

The proposed feature is to check for the failure errors when a VM deploy occurs, and if identified to be a terminal error(errors that are not transient or that wont go away with a re-try), choose another failure domain and deploy. Choosing the failure domain could be random as it is now, or can be based on resources available such as free IPs available.

This is only for worker machines, as control-plane failure domain assignment is handled by KCP.

PR #353, opened 2024-04-01
Support re-assignment of another failure domain when the machine failed to provision

jhaanvi5:
*Issue #352*

*Description of changes:* CAPC chooses a [random failure domain](https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/blob/d597e8056e82ac6418171e66fa2532b2daaef987/controllers/cloudstackmachine_controller.go#L185) to deploy worker machines. When this VM deploy fails, irrespective of the type of error, CAPC will keep re-attempting to deploy a VM until CAPI replaces the owner machine. This change introduces a failure domain balancer with fallback mechanism, primary balancer selects failure domain from network with most free IPs and if it fails it fallbacks to random failure domain selection.

*Testing performed:* 
`make test`

### DIFF
--- a/pkg/failuredomains/balancer.go
+++ b/pkg/failuredomains/balancer.go
@@ -23,11 +23,12 @@ import (
 
 	"github.com/apache/cloudstack-go/v2/cloudstack"
 	"k8s.io/utils/pointer"
-	infrav1 "sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta3"
-	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta3"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
 )
 
 type Balancer interface {
--- a/pkg/failuredomains/balancer.go
+++ b/pkg/failuredomains/balancer.go
@@ -1,0 +1,238 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package failuredomains
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+
+	"github.com/apache/cloudstack-go/v2/cloudstack"
+	"k8s.io/utils/pointer"
+	infrav1 "sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta3"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type Balancer interface {
+	Assign(ctx context.Context, csMachine *infrav1.CloudStackMachine, capiMachine *clusterv1.Machine, fds []infrav1.CloudStackFailureDomainSpec) error
+}
+
+func NewFailureDomainBalancer(csClientFactory ClientFactory) Balancer {
+	return newReassigningFailureDomainBalancer(newFallingBackFailureDomainBalancer(
+		newFreeIPValidatingFailureDomainBalancer(csClientFactory),
+		newRandomFailureDomainBalancer(),
+	))
+}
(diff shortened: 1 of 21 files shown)


pkg/failuredomains/client.go
```diff
@@ -0,0 +1,72 @@
...
+func (f *baseClientFactory) GetCloudClientAndUser(ctx context.Context, fdSpec *infrav1.CloudStackFailureDomainSpec) (csClient cloud.Client, csUser cloud.Client, err error) {
+	endpointCredentials := &corev1.Secret{}
+	key := client.ObjectKey{Name: fdSpec.ACSEndpoint.Name, Namespace: fdSpec.ACSEndpoint.Namespace}
+	if err := f.Get(ctx, key, endpointCredentials); err != nil {
+		return nil, nil, errors.Wrapf(err, "getting ACSEndpoint secret with ref: %v", fdSpec.ACSEndpoint)
+	}
+
+	clientConfig := &corev1.ConfigMap{}
+	key = client.ObjectKey{Name: cloud.ClientConfigMapName, Namespace: cloud.ClientConfigMapNamespace}
+	_ = f.Get(ctx, key, clientConfig)
```
vivek-koppuru, line 54: Add a comment here mentioning that client config is optional, hence we can ignore the error

vivek-koppuru (approved):

pkg/failuredomains/balancer.go
```diff
@@ -0,0 +1,239 @@
...
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-cloudstack/api/v1beta3"
+	"sigs.k8s.io/cluster-api-provider-cloudstack/pkg/cloud"
+)
+
+type Balancer interface {
+	Assign(ctx context.Context, csMachine *infrav1.CloudStackMachine, capiMachine *clusterv1.Machine, fds []infrav1.CloudStackFailureDomainSpec) error
```
g-gaston, line 35: What does entail to assign a failure domain? Only setting the field in the machines?

In that case I would vote to rewrite this to return the actual failure domain, and let the caller set the field in the machine. This is just a personal opinion, but I think that minimizing side effects makes things easier to understand. In this case, it separates the logic to decide a failure domain from the code to set the failure domain.

pkg/failuredomains/balancer.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func NewFailureDomainBalancer(csClientFactory ClientFactory) Balancer {
+	return newReassigningFailureDomainBalancer(newFallingBackFailureDomainBalancer(
+		newFreeIPValidatingFailureDomainBalancer(csClientFactory),
+		newRandomFailureDomainBalancer(),
+	))
```
g-gaston, line 42: I appreciate the idea of using composition here. However, what's the benefit in this case?

It doesn't seem like we are reusing any of the Balancers. Do we expect to have to use them in a different way at some point? Can we actually use them separately? in other words, are they truly independent?

If it's just to decompose the problem into smaller parts? Does that make the program simpler or more complex?

pkg/failuredomains/balancer.go
```diff
@@ -0,0 +1,239 @@
...
+	if capiMachineHasFailureDomain(capiMachine) && !csMachine.HasFailed() {
+		csMachine.Spec.FailureDomainName = *capiMachine.Spec.FailureDomain
+		assignFailureDomainLabel(csMachine, capiMachine)
+	} else if err := r.delegate.Assign(ctx, csMachine, capiMachine, fds); err != nil {
+		return err
+	}
+
+	if capiMachineHasFailureDomain(capiMachine) && csMachine.HasFailed() {
+		capiMachine.Spec.FailureDomain = pointer.String(csMachine.Spec.FailureDomainName)
+	}
```
g-gaston, line 74: I'll admit I'm not an expert on this are of the code base, but the intent here is not clear to me. If this is the best way to write this logic, I think it could benefit from some comments to explain what it's supposed to be doing (at a higher level than the code).

pkg/failuredomains/balancer.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func NewFailureDomainBalancer(csClientFactory ClientFactory) Balancer {
+	return newReassigningFailureDomainBalancer(newFallingBackFailureDomainBalancer(
+		newFreeIPValidatingFailureDomainBalancer(csClientFactory),
+		newRandomFailureDomainBalancer(),
+	))
+}
+
+type reassigningFailureDomainBalancer struct {
```
g-gaston, line 45: I think all this implementations of Balancer could use some go docs explaining what they do.

vishesh92:
@jhaanvi5 What is the status/ETA for this PR?
I am going through the open issues and tasks for the next release of CAPC (`v0.5.0`). It would be good to have an estimate to decided whether to include this in `v0.5.0` or not.

cc: @g-gaston @vignesh-goutham @vivek-koppuru

vivek-koppuru:
@vishesh92 Figuring this out, I think this is done for the most part so it would be great to get your review as well on this. This came as an ask so need to follow up based on whether any of these changed can't make it to the v0.5.0 release, but we would love to have it included pending review from you all!

vishesh92:
> @vishesh92 Figuring this out, I think this is done for the most part so it would be great to get your review as well on this. This came as an ask so need to follow up based on whether any of these changed can't make it to the v0.5.0 release, but we would love to have it included pending review from you all!

@vivek-koppuru I reviewed this PR a little bit and have same questions as @g-gaston . Could you please resolve/address the comments? This PR also has some conflicts now.

PR #328, opened 2023-12-06
DiskOffering in CloudStackMachine CRD should be a pointer to CloudStackResourceDiskOffering

hrak:
*Issue #, if available:*

#326 

*Description of changes:*

This is a breaking change in the (unreleased) v1beta3 API which changes the `DiskOffering` property of `CloudStackMachine` to be a pointer to `CloudStackResourceDiskOffering`

Reasoning is described in the above linked issue.

*Testing performed:*

make test

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #326, opened 2023-12-06
DiskOffering in CloudStackMachine CRD should be a pointer to CloudStackResourceDiskOffering


hrak:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

`DiskOffering` is an optional field in the `CloudStackMachine` CRD. It is currently defined as:

```
	// CloudStack disk offering to use.
	// +optional
	DiskOffering CloudStackResourceDiskOffering `json:"diskOffering,omitempty"`
```

This can lead to unnecessary reconciliations or even unintended behavior when this field is omitted from a yaml definition, because the entire structure will be included in the final object definition with its default values:

```yaml
spec:
  diskOffering:
    customSizeInGB: 0
    device: ""
    filesystem: ""
    label: ""
    mountPath: ""
  failureDomainName: failure-domain-1
  offering:
    name: Medium Instance (8 GB / 4 CPU)
  sshKey: h.rakers-ed25519
```

`DiskOffering` should be changed to `*CloudStackResourceDiskOffering` to comply with the [Kubernetes API conventions](https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#optional-vs-required).

PR #359, opened 2024-05-24
DiskOffering in CloudStackMachine CRD should be a pointer to CloudStackResourceDiskOffering

vishesh92:
Recreated PR https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/328 by @hrak  with minor fixups

*Issue #, if available:*

#326 

*Description of changes:*

This is a breaking change in the (unreleased) v1beta3 API which changes the `DiskOffering` property of `CloudStackMachine` to be a pointer to `CloudStackResourceDiskOffering`

Reasoning is described in the above linked issue.

*Testing performed:*

make test


g-gaston (commented):
It's possible that I missed something, but isn't v1beta3 already released? If this is a breaking change, don't we need v1beta4?

vishesh92:
> It's possible that I missed something, but isn't v1beta3 already released? If this is a breaking change, don't we need v1beta4?

As per my understanding, this will allow users migrate to `v1beta3` without any loss in metadata when they are upgrading CAPC.

g-gaston (commented):
lgtm
@vignesh-goutham for final approval since he had more context on v1beta3

vignesh-goutham:
> > It's possible that I missed something, but isn't v1beta3 already released? If this is a breaking change, don't we need v1beta4?
> 
> As per my understanding, this will allow users migrate to `v1beta3` without any loss in metadata when they are upgrading CAPC.

Could you help me understand how this will work? Looking at the code, there are conversions to and from v1beta3 to other versions. An existing CloudStackMachine on v1beta3 cant mutate to use a pointer to diskOffering. Its recommended to do only additive changes to released APIs, this seems like it might not be additive in nature, unless I'm missing something.

vishesh92:
@vignesh-goutham let me know if we can merge this?

PR #448, opened 2025-06-17
Add multiple nics support to nodes

harikrishna-patnala:
*Description of changes:*
This PR adds the capability for attaching multiple networks to the VM in k8s.

Currently 

*Testing performed:*


api/v1beta3/cloudstackmachine_types.go
```diff
@@ -91,6 +99,12 @@ type CloudStackMachineSpec struct {
 	UncompressedUserData *bool `json:"uncompressedUserData,omitempty"`
 }
 
+type NetworkSpec struct {
+	Name string `json:"name"`
```
vishesh92, line 103: Add `Id` so that the user can pass in the id as well. Update the id in the kubernetes spec as well.
This id will be further used while deploying the vm with additional networks.

harikrishna-patnala: added in the recent commit @vishesh92 , thank you

vishesh92 (approved):
/lgtm

PR #453, opened 2025-07-07
Added validations and templates for multiple networks configuration

harikrishna-patnala:
*Description of changes:*
This PR related to https://github.com/kubernetes-sigs/cluster-api-provider-cloudstack/pull/448, here we are adding multiple validations, csMachine status updates for the network and IPs, and a template for explaining configuration for multiple nics.

*Testing performed:*


templates/cluster-template-multiple-networks.yaml
```diff
@@ -0,0 +1,140 @@
...
+      offering:
+        name: ${CLOUDSTACK_WORKER_MACHINE_OFFERING}
+      template:
+        name: ${CLOUDSTACK_TEMPLATE_NAME}
+      networks:
+        - name: ${CLOUDSTACK_NETWORK_NAME} # default primary network and should match with failureDomains.zone.network.name
+          ip: 10.1.1.41 # optional ip address in the network
+        - name: # extra network, it can be network name or ID
+          ip: 10.1.1.51 # optional ip address in the network
+        - id: # extra network, it can be network name or ID
```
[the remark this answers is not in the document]
vishesh92: It should be possible to pass the name of the extra network from environment variable

pkg/cloud/instance.go
```diff
@@ -302,38 +312,163 @@ func (c *client) CheckLimits(
...
-			entry["ip"] = net.IP
-		}
+
 		ipToNetworkList = append(ipToNetworkList, entry)
 	}
+
 	return ipToNetworkList, nil
 }
 
+func (c *client) resolveNetworkReference(net infrav1.NetworkSpec) (string, *cloudstack.Network, error) {
```
vishesh92, line 434:
```suggestion
func (c *client) resolveNetwork(net infrav1.NetworkSpec) (string, *cloudstack.Network, error) {
```
to maintain consistency

pkg/cloud/instance.go
```diff
@@ -302,38 +312,163 @@ func (c *client) CheckLimits(
...
+			return "", nil, err
+		}
+		return resolvedNet.Id, resolvedNet, nil
+	}
+
+	resolvedNet, _, err := c.cs.Network.GetNetworkByID(net.ID, cloudstack.WithProject(c.user.Project.ID))
+	if err != nil {
+		return "", nil, errors.Wrapf(err, "failed to get network %q by ID", net.ID)
+	}
+	return net.ID, resolvedNet, nil
```
vishesh92, line 447: resolvedNet.ID & net.ID is always same. I don't think we need to return both net.ID & resolvedNet

pkg/cloud/instance.go
```diff
@@ -302,38 +312,163 @@ func (c *client) CheckLimits(
...
+		}
+		if !freeIPExists {
+			return nil, errors.Errorf("no free IPs available in network %q", resolvedNet.Id)
+		}
+	}
+
+	return map[string]string{
+		"networkid": resolvedNet.Id,
+	}, nil
+}
```
vishesh92, line 393: I think we can merge these 4 methods.

pkg/cloud/instance.go
```diff
@@ -302,38 +312,129 @@ func (c *client) CheckLimits(
...
+
+	resolvedNet, _, err := c.cs.Network.GetNetworkByID(net.ID, cloudstack.WithProject(c.user.Project.ID))
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get network %q by ID", net.ID)
+	}
+	return resolvedNet, nil
+}
+
+func (c *client) validateIPInCIDR(ipStr string, net *cloudstack.Network, netID string) error {
+	ip := netpkg.ParseIP(ipStr)
```
vishesh92, line 421: We can make this method simpler by passing in only the CIDR and IP to check. That will make this method useful for other scenarios as well.

templates/cluster-template-multiple-networks.yaml
```diff
@@ -0,0 +1,148 @@
...
+metadata:
+  name: ${CLUSTER_NAME}-control-plane
+spec:
+  template:
+    spec:
+      sshKey: ${CLOUDSTACK_SSH_KEY_NAME}
+      offering:
+        name: ${CLOUDSTACK_CONTROL_PLANE_MACHINE_OFFERING}
+      networks:
+        - name: ${CLOUDSTACK_NETWORK_NAME} # (optional) default primary network; must match failureDomains.zone.network.name
```
vishesh92, line 77: Not sure if the comments are in the generated spec or not. Check if they are part of the generated spec or not.
It would be better to allow at least specifying one more network name here.

harikrishna-patnala: agreed with you Vishesh, I'm taking out the yaml all together as its not adding good value on the default template.

vishesh92 (approved):

PR #485, opened 2026-05-20
Deregister control plane VM from Kubernetes_API_Server LB rule on CloudStackMachine delete

metbog:
*Issue #, if available:*

N/A

*Description of changes:*

Currently, when a control plane `CloudStackMachine` is deleted, CAPC only destroys the underlying CloudStack VM - it never detaches the VM from the `Kubernetes_API_Server` load balancer rule. VM removal can take a while, and the CloudStack LB health check only marks the member offline after some delay, so API traffic keeps hitting the about-to-be-expunged kube-apiserver throughout that window. This causes stale members and intermittent control plane reachability issues during scale-down or rolling replace (most visibly during KCP-driven control plane rollouts)

This change mirrors the CAPO (`cluster-api-provider-openstack`) `DeleteLoadBalancerMember` pattern, with one addition: removal is attempted **early** - as soon as the parent CAPI Machine is marked for deletion - instead of only on the `CloudStackMachine` delete path. The late removal is kept as a safety net

### Changes

- **`pkg/cloud/isolated_network.go`** - added `RemoveVMFromLoadBalancerRule` to the `IsoNetworkIface`. Idempotent: returns `nil` when `LBRuleID` is empty or the VM is not currently a member of the rule. Otherwise calls CloudStack `removeFromLoadBalancerRule`

- **`controllers/cloudstackmachine_controller.go`** - two LB-detach stages plus a small helper:
  - `usesAPIServerLB()` - extracted predicate shared by all LB stages: non-routed isolated network + populated `IsoNet` + `InstanceID` set
  - `RemoveFromLBIfMachineDeleting` - new stage in the regular `Reconcile()` pipeline, ordered **before** `AddToLBIfNeeded`. Fires when the parent CAPI Machine has a `DeletionTimestamp`, the machine is a control plane, and `usesAPIServerLB()` is true. Detaches the VM from the LB rule promptly, before KCP starts cordon/drain. `AddToLBIfNeeded` has a matching guard so a deleting Machine never gets re-added in the same loop
  - `RemoveFromLBIfNeeded` - new stage called from `ReconcileDelete` immediately before `DestroyVMInstance`, acting as a safety net for the case where the early step did not run (CSMachine deleted directly, controller downtime, etc.). Two implementation notes:
    - Detects control plane via the `clusterv1.MachineControlPlaneLabel` on the `CloudStackMachine` itself, because on the delete path the common stages skip loading the CAPI parent Machine (`RunIf(... DeletionTimestamp().IsZero() ..., GetParent(...))`).
    - Fetches the `CloudStackIsolatedNetwork` via `GetObjectByName(IsoNetMetaName(...), r.IsoNet)`, since the delete path does not pre-load it (unlike the regular `Reconcile` path)
  - **CAPI Machine watch predicate widened** - the existing `clusterv1.Machine` update predicate only enqueued the owned `CloudStackMachine` when `Bootstrap.DataSecretName` went from `nil` to set. It now also enqueues when `DeletionTimestamp` transitions from zero to non-zero. Without this, the early-removal stage would only run on the next periodic resync (1m default) - too late.

- **`pkg/cloud/isolated_network_test.go`** - 5 new unit tests covering: empty `LBRuleID` no-op, VM not a member no-op, happy-path removal, list error, remove error

- **`pkg/mocks/mock_client.go`** - regenerated via `make generate-mocks` (gitignored; auto-rebuilt by `make test`)

- **`docs/book/src/topics/cloudstack-permissions.md`** - added the required `removeFromLoadBalancerRule` permission

### Behavior matrix

| Trigger | Machine role | FD network | RoutingMode | Result |
|---|---|---|---|---|
| Parent CAPI Machine `DeletionTimestamp` set | Control plane | Isolated | `""` (non-routed) | Early LB detach during regular `Reconcile()`, before drain |
| `CloudStackMachine` deletion (`ReconcileDelete`) | Control plane | Isolated | `""` (non-routed) | Safety-net LB detach, then VM destroyed |
| Either trigger | Control plane | Isolated | non-empty (routed) | No LB call (no LB used) |
| Either trigger | Control plane | Shared / other | - | No LB call |
| Either trigger | Worker | any | - | No LB call |
| Either trigger | Any | `IsoNet` CR already gone / `LBRuleID` empty | - | Cloud-layer no-op (idempotent) |
| Either trigger | Any | VM not a member of the rule | - | Cloud-layer no-op (idempotent) |

The cluster-scoped LB rule itself continues to be cleaned up through `DisposeIsoNetResources` on cluster teardown (unchanged) - only members are removed on machine delete, matching the CAPO model

### Trade-offs / why both stages

The early stage (`RemoveFromLBIfMachineDeleting`) is what actually shortens the bad window during KCP rollouts. The late stage (`RemoveFromLBIfNeeded` in `ReconcileDelete`) is intentionally kept because:
- The controller may have been down when the parent Machine was marked for deletion
- A user can delete the `CloudStackMachine` directly, bypassing the CAPI Machine
- It costs at most one extra `ListLoadBalancerRuleInstances` call (idempotent no-op when already detached)

*Testing performed:*

- `go build ./...` - clean
- `go vet ./controllers/... ./pkg/cloud/...` - clean
- `REPO_ROOT=$(pwd) go run github.com/onsi/ginkgo/v2/ginkgo --label-filter='!integ' ./pkg/cloud` - **137 passed, 0 failed**, 24 integration specs skipped as expected. All 5 new specs green
- Manually verified the regenerated mock surface includes `RemoveVMFromLoadBalancerRule` on `MockClient`

### DIFF
--- a/controllers/cloudstackmachine_controller.go
+++ b/controllers/cloudstackmachine_controller.go
@@ -126,6 +126,7 @@ func (r *CloudStackMachineReconciliationRunner) Reconcile() (retRes ctrl.Result,
 		r.ConsiderAffinity,
 		r.GetOrCreateVMInstance,
 		r.RequeueIfInstanceNotRunning,
+		r.RemoveFromLBIfMachineDeleting,
 		r.AddToLBIfNeeded,
 		r.GetOrCreateMachineStateChecker,
 	)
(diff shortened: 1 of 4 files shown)


controllers/cloudstackmachine_controller.go
```diff
@@ -292,8 +293,45 @@ func (r *CloudStackMachineReconciliationRunner) RequeueIfInstanceNotRunning() (r
 	return ctrl.Result{}, nil
 }
 
+// usesAPIServerLB reports whether the CloudStackMachine participates in the API server
+// load balancer rule: it is on a non-routed isolated network and has an InstanceID.
+// Caller is responsible for ensuring r.IsoNet is populated.
+func (r *CloudStackMachineReconciliationRunner) usesAPIServerLB() bool {
```
Damans227, line 299: The description says usesAPIServerLB is shared by all the LB stages, but AddToLBIfNeeded still checks the same things itself. Worth using the helper there too so there is one place that decides this.

pkg/cloud/isolated_network_test.go
```diff
@@ -423,6 +423,66 @@ var _ = ginkgo.Describe("Network", func() {
 		})
 	})
 
+	ginkgo.Context("Remove VM from Load Balancer rule", func() {
```
Damans227, line 426: Tests only cover the cloud client. The two new controller stages and the watch predicate change have none. controllers/cloudstackmachine_controller_test.go already has envtest set up.

controllers/cloudstackmachine_controller.go
```diff
@@ -292,8 +293,45 @@ func (r *CloudStackMachineReconciliationRunner) RequeueIfInstanceNotRunning() (r
...
+		r.ReconciliationSubject.Spec.InstanceID != nil
+}
+
+// RemoveFromLBIfMachineDeleting detaches the VM from the API server load balancer rule as soon
+// as the parent CAPI Machine is marked for deletion, before CAPI deletes the CloudStackMachine.
+// VM removal can take a while, and the CloudStack LB health check only marks the member offline
+// after some delay, so without this step API traffic keeps hitting the about-to-be-replaced
+// kube-apiserver throughout that window. RemoveFromLBIfNeeded (in ReconcileDelete) remains as a
+// safety net.
+func (r *CloudStackMachineReconciliationRunner) RemoveFromLBIfMachineDeleting() (retRes ctrl.Result, reterr error) {
```
Damans227, line 312: Tried this with a single control plane node? Deleting that one Machine takes the last member out of the LB while it's still serving, so drain has nothing to talk to. Probably rare, but worth a check.

Pearl1594:
@metbog can you please address the comments, so we can test/merge the PR. Thanks

